PR #1, opened 2019-03-27
Replace Browserify with Webpack

MatthewGidcomb:


web-portal/build/build.js
```diff
@@ -0,0 +1,23 @@
+/**
```
chriswininger: might consider calling this prod.js since the other is called dev.js

web-portal/build/utils.js
```diff
@@ -0,0 +1,40 @@
...
+      return ExtractTextPlugin.extract({
+        use: loaders,
+        fallback: 'vue-style-loader'
+      })
+    } else {
+      return ['vue-style-loader'].concat(loaders)
+    }
+  }
+
+  return {
```
chriswininger: why call this twice, won't it give the same thing both times

MatthewGidcomb: This was copied from the Vue CLI 2 config. I think the return value is passed to code that only uses one of the fields on the returned object.
I may have also removed some code that handled other preprocessors.

PR #20, opened 2019-11-13
Don't Store Token In Global State

chriswininger:
Setting the token on the default axios headers in dangerous when your
code can potentially be ran server side. The axios default headers are
global to all user sessions

Instead these must be accessed from the auth provider in components and
threaded to the correct place as needed


web-portal/store/index.js
```diff
@@ -173,6 +173,7 @@ export const mutations = {
 
 export const actions = {
   nuxtServerInit: (ctx, { req }) => {
+    debugger
```
MatthewGidcomb: Should probably remove this before merging it.

chriswininger: done

chriswininger: thx

PR #67, opened 2019-12-15
Widget skeleton

Digi-D:
Basic widget demo that pulls info from our API server.


widget/src/apiCaller.js
```diff
@@ -0,0 +1,20 @@
+// Options:
+// local
+// staging - https://staging-api.infinite.industries/events/current/verified/
+// production - https://api.infinite.industries/events/current/verified/
+
+// Create methods ^%$#^%$#!!!!
+
+
+export function APICaller(callback){
```
MatthewGidcomb: If this used `export default function`... instead then it could be imported without the extra brackets, e.g. `import APICaller from './apiCaller.js'` instead of `import { APICaller }`...

Digi-D:
No longer really resembling a "skeleton" ;) Merging and opening a new branch that will concentrate more on the design. Need to review the testing strategy as I am not good at that stuff.

widget/src/apiCaller.js
Digi-D: Done.

PR #72, opened 2019-12-25
Widget skeleton

Digi-D:
Add the rest of the widget components that did not get added with the last pull request


Digi-D:
Ooops. Adding all the bits that didn't come through with the last pull request :|

PR #83, opened 2020-01-09
Add Winston logging to portal

MatthewGidcomb:
- Adds a logger utility based on the API server's
- Replaces console logging in the internal-api endpoints with the utility, adding a few more for completeness

### DIFF
--- a/web-portal/internal-api/utils.js
+++ b/web-portal/internal-api/utils.js
@@ -1,0 +1,34 @@
+/**
+ * Utilities for the internal API endpoints -- don't use these in SSR code
+ */
+
+// logging support (copied from api-server)
+const winston = require('winston')
+const { createLogger, format, transports } = winston
+const { combine, timestamp, label, printf, colorize } = format
+const env = process.env.ENV || 'dev'
+const myFormat = printf(({ level, message, label, timestamp }) => {
+  return `${level}: ${timestamp} [${label}] -- ${message}`
+})
+
+const labelOpt = { label: env + ':web-portal:' + process.pid }
+
+const logger = createLogger({
+  format: combine(
+    label(labelOpt),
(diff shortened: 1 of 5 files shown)


web-portal/internal-api/utils.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Utilities for the internal API endpoints -- don't use these in SSR code
+ */
+
+// logging support (copied from api-server)
+const winston = require('winston')
+const { createLogger, format, transports } = winston
+const { combine, timestamp, label, printf, colorize } = format
+const env = process.env.ENV || 'dev'
```
MatthewGidcomb, line 9: Does this need to be added to the `.env` file for the web portal?

chriswininger: yeah that should be in there so you can log the environment. It's already there on the web portal servers

chriswininger: if you don't have it set locally it will default to dev, which is what you would want anyyways

web-portal/package.json
```diff
@@ -32,7 +32,9 @@
     "vue-clipboard2": "^0.3.0",
     "vue-moment": "^4.0.0",
     "vue-smoothscroll": "^0.2.0",
-    "vue2-editor": "^2.10.0"
+    "vue2-editor": "^2.10.0",
+    "winston": "^3.2.1",
+    "winston-daily-rotate-file": "^4.4.1"
```
chriswininger: we don't actually need this file. I probably added that to mine by accident. that's something we might use in the future

MatthewGidcomb: Just pushed a commit to remove `winston-daily-rotate-file`. Any objections to merging the PR?

PR #92, opened 2020-01-19
Refactor widget to support runtime configuration

MatthewGidcomb:
Site and API URLs are now defined in a ConfigService and chosen based on a data attribute from the element the widget is inserted into.


widget/src/configService.js
```diff
@@ -0,0 +1,53 @@
...
+
+    static getSiteUrl() {
+        return URLS[mode].SITE
+    }
+
+    static getApiUrl() {
+        return URLS[mode].API
+    }
+}
+
```
MatthewGidcomb, line 53: An alternative would be to export individual `const`s from this file, and import them directly instead of the class.

E.g.,

```js
import { API_URL, SITE_URL } from './configService.js' // possibly even rename 'config.js'
console.log(API_URL) // etc.
``

PR #126, opened 2020-03-29
Add a Remote Page

chriswininger:
Added a page which displays only remote events. This could be a place to
send people during the pandemic

### DIFF
--- a/web-portal/store/index.js
+++ b/web-portal/store/index.js
@@ -50,6 +50,12 @@ export const getters = {
     return state.all_local_events
   },
 
+  GetAllRemoteEvents: (state) => {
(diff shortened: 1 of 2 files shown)


web-portal/store/index.js
```diff
@@ -50,6 +50,12 @@ export const getters = {
     return state.all_local_events
   },
 
+  GetAllRemoteEvents: (state) => {
+    return state.all_local_events.filter((localEvent) => {
+      return localEvent.tags && localEvent.tags.indexOf('remote') >= 0
```
MatthewGidcomb, line 55: I've been using `includes` in the Submission Form's tags handling.

PR #172, opened 2020-09-07
Caw adding nest api

chriswininger:
An initial commit adding the newly refactored api into the module. This version leverages the nest framework


api-server-nest/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "infinite-api",
+  "version": "0.0.1",
+  "description": "",
+  "author": "",
+  "private": true,
+  "license": "UNLICENSED",
```
MatthewGidcomb, line 7: Should probably change this to `"MIT"` in case we end up copying this directory to `api-server` like we did with the Nuxt refactor.

api-server-nest/src/app.module.ts
```diff
@@ -0,0 +1,27 @@
...
+require('dotenv').config();
+
+@Module({
+    imports: [
+        SequelizeModule.forRoot({
+            dialect: 'postgres',
+            autoLoadModels: true,
+            synchronize: true,
+            host: process.env.DB_HOST,
+            port: +process.env.DB_PORT,
```
MatthewGidcomb, line 23: Is the `+` here to trigger a cast to `Number`? If so, seems like `parseInt` would be clearer.

api-server-nest/src/current-events/current-events.controller.ts
```diff
@@ -0,0 +1,89 @@
...
+import {CurrentEventsService} from "./current-events.service";
+import {CurrentEvent} from "./dto/current-event.model";
+import {Venue} from "../venues/dto/venue.model";
+import {FindOptions} from "sequelize";
+import {AuthGuard} from "../authentication/auth.guard";
+import {ApiBearerAuth, ApiOperation, ApiResponse, ApiTags} from "@nestjs/swagger";
+import {VERSION_1_URI} from "../utils/versionts";
+import {LoggingInterceptor} from "../logging/logging.interceptor";
+
+type EmbedableModels = typeof Venue
```
MatthewGidcomb, line 16: Apologies for pedantry but I think this should be `EmbeddableModels`

api-server-nest/src/current-events/dto/current-event.model.ts
```diff
@@ -0,0 +1,127 @@
...
+    @Column
+    @ApiProperty({example: 'infinite-gallery-opening'})
+    slug: string;
+
+    @Column
+    @ApiProperty({example: false})
+    multi_day: boolean;
+
+    @Column
+    @ApiProperty({example: 'f467e7a0-a066-11ea-aa51-cdc3fe7afefa'})
```
MatthewGidcomb: Should this look more like line 40?

api-server-nest/src/current-events/dto/current-event.model.ts
```diff
@@ -0,0 +1,127 @@
...
+    @ApiProperty({example: EXAMPLE_DATE})
+    createdAt: Date;
+
+    @Column
+    @ApiProperty({example: EXAMPLE_DATE})
+    updatedAt: Date;
+
+    @Column
+    @ApiProperty({example: 'radio-mc-radio-station'})
+    reviewed_by_org: boolean;
```
MatthewGidcomb: I think this is a string and not a boolean.

PR #175, opened 2020-09-13
New read me

Digi-D:
Quick Readme Update


MatthewGidcomb (approved):

README.md
```diff
@@ -1,10 +1,41 @@
-# Infinite Industries 
+# Infinite Industries
+<img src="./docs/doc-splash.png">
 
 ## Mission Statement
+<p>Infinite Industries makes experimental contemporary culture more accessible to everyone. Our digital platform and open-source tools highlight cultural events and experiences so more of us can discover, share and participate in them.
```
chriswininger: you don't really need these <p> tags here. Just a new line will do the trick

README.md
```diff
@@ -1,10 +1,41 @@
...
+</p>
 
-## Directory Structure
+<p>
+<b>why the name?</b>
```
chriswininger: Instead of <b> you do wrap this text with **Why the name?**

It will do the same thing

Also you can ditch the paragraphs and just use new lines

README.md
```diff
@@ -1,10 +1,41 @@
...
+Artists constantly evolve, inventing new genres and redefining their methods. The processes and mediums of the past aren’t always sufficient for the challenges of the future. We can thank artists for an infinity of possibilities. Production trends organize beneath the umbrella term "industry," creating an infinity of industries and our special name ;)
+</p>
```
chriswininger: So that's what our name means :-) 👍

README.md
```diff
@@ -1,10 +1,41 @@
...
+<p>Our efforts are currently focused on Lexington, Kentucky USA with lots of future plans for world domination.</p>
 
+## Project Components
+The project is split into API and various components that source information from it. We are intentionally trying to centralize the information and de-centralize the types of services that consume the data.
+### API
```
chriswininger, line 20: Might want to add a new line between end of this paragraph and the next header. It's also typicial in markdown to leave a blank line between the header and the body text so for example

## Foo Bar

Some text

## Much Foo

More text

README.md
```diff
@@ -1,10 +1,41 @@
-# Infinite Industries 
+# Infinite Industries
+<img src="./docs/doc-splash.png">
```
chriswininger: If you want to do this in markdown as well you can do My Alt Text

In the case of images I'm not sure it matters as much which way we do it because neither is exactly super readable, but might be nice to stay consistent

README.md
```diff
@@ -1,10 +1,41 @@
...
+<p>
+<b>why the name?</b>
+Artists constantly evolve, inventing new genres and redefining their methods. The processes and mediums of the past aren’t always sufficient for the challenges of the future. We can thank artists for an infinity of possibilities. Production trends organize beneath the umbrella term "industry," creating an infinity of industries and our special name ;)
+</p>
+<p>Our efforts are currently focused on Lexington, Kentucky USA with lots of future plans for world domination.</p>
 
+## Project Components
+The project is split into API and various components that source information from it. We are intentionally trying to centralize the information and de-centralize the types of services that consume the data.
+### API
+Our db and a RESTful interface to our db.
```
chriswininger: It's not exactly the db. I would maybe say something like:

A backend service which brokers access to the database

README.md
```diff
@@ -1,10 +1,41 @@
...
+
+[API Documentation](https://github.com/infinite-industries/infinite/tree/master/api-server)
+### Web Portal
+Most of the world sees us here.
```
chriswininger: We could potentially call this out a bit more explicitly:

The frontend web application served at [infinite.industries](https://infinite.industries). This is how most of the world sees us.

chriswininger (commented):
Awesome! Thanks for cleaning this up.

The wording looks good. I made a few suggestions.

I believe we spoke about this on the call but I also pointed out places where we were mixing in html into the markdown.

In particular you don't need those <p> tags. You can achieve the same things with just normal new lines

PR #183, opened 2020-12-06
Add tag filtering support for events and current-events

MatthewGidcomb:


web-portal/store/admin.js
```diff
@@ -41,7 +41,7 @@ export const actions = {
 
     ApiService.all([
       ApiService.get(CURRENT_EVENTS_NON_VERIFIED_PATH, idToken),
-      ApiService.get('/events/non-verified/tags/online-resource', idToken)
+      ApiService.get('/v1/events/non-verified?tags=online-resource', idToken)
```
MatthewGidcomb: I didn't do it in this pass, but we should consider building the version segment into the API service.

chriswininger: I'm confused. I thought I did that: `@Controller(`${VERSION_1_URI}/events`)`

api-server-nest/src/utils/get-common-query-terms-for-events.ts
```diff
@@ -0,0 +1,21 @@
+import { Op, WhereOptions } from "sequelize";
+// import {HttpException, HttpStatus} from "@nestjs/common";
+
+// TODO: it would be nice to be able to say tags=-tag to filter on events _without_ tag
+/**
+ * Build query options for events (or current events)
+ * 
+ * @param verified true for only verified events, false for only unverified events, null for both
+ * @param tags tag or array of tags to filter by (when using an array, only events with ALL tags are returned)
```
chriswininger, line 9: FYI, one of the nice things about typescript is you can skip this kind of documentation.

I'd argue the function name and type decorations speak for them selves; though you could perhaps change the name to buildQueryOptionsForEventServiceCalls? I'm fine with it either way and definitely, not insisting you delete the comments, but I've been reading Clean Code and they make the case that comments outside of public apis just get in the way and become stale and are indicators that you need better function names or type definitions.

api-server-nest/src/events/events.controller.ts
```diff
@@ -65,8 +72,14 @@ export class EventsController {
...
     @ApiBearerAuth()
-    getAllCurrent(@Query('embed') embed: string[] | string = []): Promise<Event []> {
-        const findOptions = getOptionsForEventsServiceFromEmbedsQueryParam(embed);
+    getAllCurrent(
+        @Query('embed') embed: string[] | string = [],
+        @Query('tags') tags: string[] | string = [],
+    ): Promise<Event []> {
+        const findOptions = {
+            ...getOptionsForEventsServiceFromEmbedsQueryParam(embed),
+            where: getCommonQueryTermsForEvents(null, tags)
```
chriswininger, line 81: nice job consolidating this

web-portal/store/admin.js
chriswininger: yes good idea.That should be part of the base url, so we don't have to update it everywhere. Honestly I think we can just do this in the api as part of the base url we already have can't we?

web-portal/store/index.js
```diff
@@ -1,7 +1,7 @@
 import { ApiService } from '../services/ApiService'
 import { getEmptyCalendarEvent } from '../services/ResourceTemplateService'
 
-const CURRENT_EVENTS_VERIFIED_PATH = '/current-events/verified'
+const CURRENT_EVENTS_VERIFIED_PATH = '/v1/current-events/verified'
```
chriswininger: should the /v1/ part not just be set in the api url in the env?

chriswininger (approved):
Looks good to me. Left a few small thoughts as we discussed but I see no problems. Thanks for the good work :-)

web-portal/store/index.js
MatthewGidcomb: Yep. I hadn't updated my `web-portal/.env` file.

PR #188, opened 2020-12-13
Implementing current/user endpoint

chriswininger:
this adds support for getting meta data about the current user from the
server


api-server-nest/src/users/dto/user-info-resp.ts
```diff
@@ -0,0 +1,30 @@
+import {ApiProperty} from "@nestjs/swagger";
+import isNotNullOrUndefined from "../../utils/is-not-null-or-undefined";
+import cloneAttributes from "../../utils/colone-attributes";
```
MatthewGidcomb: We should fix the typo in this file name at some point.

MatthewGidcomb (approved):

PR #194, opened 2021-01-24
Filter organizer contact

chriswininger:
This adds filtering of contact organizer to all non authenticated endpoints
A new endpoint which requires authentication was created for getting an unfiltered event
The frontend was pointed to the new endpoint
Some refactoring of the existing filter code was done, to clean things up


api-server-nest/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
```diff
@@ -0,0 +1,52 @@
...
+import {Model} from "sequelize-typescript";
+
+type GenericEvent = Event | CurrentEvent
+type GenericEventList = Event[] | CurrentEvent[]
+
+export async function removeSensitiveDataForNonAdmins(
+    request: Request,
+    events: GenericEvent | GenericEventList
+): Promise<GenericEvent | GenericEventList> {
+    const isAdmin = await isAdminUser(request);
```
chriswininger, line 14: This bit here is probably not needed for the current implementation, but I'm a little worries about removing it in this pass.

It seems that this was written such that we would just apply this filter to all event endpoints and then skip it for non-admins, more like how the other api worked

MatthewGidcomb (approved):

PR #195, opened 2021-01-31
Handle Updates/Verify

chriswininger:
This finalizes work to make updates and event verification work against
the new api

* Moved update and verify to authenticated controller (This just looks
  nicer)

* Added logic to event update endpoint to correctly transform date times
  from the client as is done on the create end point

* Corrected a bug in the utility method that allows dtos to copy
  attributes in their constructor (it was transforming null to empty
  objects)

* Updated client to call the endpoints behind /authenticated

* Removed the client expectation for a response wrapper. I don't think
  this really makes sense for an update or create response. It makes the
  most sense for requests that return lists. More thought should be
  given to nomralizing how we do this in the future.


api-server-nest/src/utils/clone-attributes.ts
```diff
@@ -2,7 +2,13 @@ export default function cloneAttributes<T>(from: Partial<T>, to: T): void {
...
         const value = from[key]
 
-        if (Array.isArray(value)) {
+        if (value === undefined) {
+           return // skip
+        }
+
+        if (value === null) {
+            to[key] = null
+        } else if (Array.isArray(value)) {
```
MatthewGidcomb, line 11: Is there a particular reason to write this util ourselves vs use one from Underscore/Lodash?

MatthewGidcomb (approved):

api-server-nest/src/utils/clone-attributes.ts
chriswininger: I got scared away from lodash. I've heard folks at work complain that they don't patch security vulns and have been hostile to attempts to patch major problems, I don't know the details but comes from people I trust with a record of contributing to open source projects (though now I'm not sure if it was lodash or underscore) :shrug:

In any event it's a fair point, not good to reinvent the wheel. We should probably look into that. The good news is that by having it in our little wrapper like this we could just swap it out for lodash under the hood at some point

PR #207, opened 2021-02-20
Add winston logging to Nest project

MatthewGidcomb:
- Adds nest-winston module to bring in prebuilt Nest module for winston
- Injects it into the app and overrides Nest logger with it for consistent logging experience
- Injects logger into events controller and bitly service as examples


chriswininger (approved):
looks good to me

PR #209, opened 2021-02-21
 issue-201/issue-208: Fix Venue Submission

chriswininger:
* slugs for new venues are generated server side and not marked required
  for posting
 
* fetching a venue by id returns the expected response format to the
  client

* unrelated quick fix to get e2e specs working again after addition of Winston Module, there's likely a better way to do this that does not involve so much duplication of code, but alas here are for now


api-server-nest/src/venues/dto/single-venue-response.ts
```diff
@@ -0,0 +1,15 @@
+import {ResponseWrapper} from "../../dto/response-wrapper";
+import isNotNullOrUndefined from "../../utils/is-not-null-or-undefined";
+import {VenueModel} from "../models/venue.model";
+
+export class SingleVenueResponse extends ResponseWrapper {
+    venue: VenueModel
+
+    constructor(copy?: Partial<SingleVenueResponse>) {
+        super(copy)
```
chriswininger, line 9: I opted to break with the pattern of using that clone utility and just to this a bit more manually, it was causing problems again and is probably over kill for this use case where we are likely never to have more than one or two fields to copy

MatthewGidcomb (approved):

PR #219, opened 2021-03-19
Fix issue with server dropping event description

MatthewGidcomb:
Not sure why the IsOptional decorator is necessary, but without it the description is silently dropped.


MatthewGidcomb:
Superseded by #221

PR #234, opened 2021-05-30
Release

MatthewGidcomb:
- Throttle the jumbotron so it only appears once per browsing session
- Various file cleanups


chriswininger (approved):
+1

Issue #230, opened 2021-04-18
Soft-delete for Venues


MatthewGidcomb:
The primary benefit would be that we could hide venues from the public-facing list used on the submission form without actually deleting them from the database and thus leaving bad reverences on old events. Reasons for hiding venues include closing down (RIP Cosmic Charlie's) or changing name / address, where we might want to create a new venue so that older events still refer to a venue with venue details as they were when the event occurred.

That said, maybe instead of a classic soft-delete we want something functionally similar like an "active" flag?

MatthewGidcomb:
Implemented in #235

PR #235, opened 2021-05-30
 issue #230: Venue Update Capabilities

chriswininger:
This adds the ability to manage venues.

This means that admins can soft delete a venue or update it

There is a flag which can be used to return only non-deleted venues

Tests were added for all venue endpoints

The e2e specs were updated to log output from server under test to a
separate file to simplify debugging

Log files are truncated between test runs


api-server/migrations/20210509145547-add-soft-delete.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      ALTER TABLE venues
+        ADD COLUMN is_soft_deleted BOOLEAN
```
MatthewGidcomb, line 7: Does Sequelize not provide drop-in /opt-in support for soft deletes?

MatthewGidcomb (approved):
Looks good.

api-server/migrations/20210509145547-add-soft-delete.js
chriswininger: nah, I doubt it. It's not that tightly integrated. It would need control of your schema for that, but we control the schema through simple migrations.

PR #240, opened 2021-07-12
Accessibility

rlmead:
**new db migration file**

⬆️ up
- adds `accessibility` column to `events` table
- drops and replaces `current_events` view to include `accessibility` column

⬇️ down
- drops `current_events` view
- drops `accessibility` column from `events` table
- replaces `current_events` view with its former version, without `accessibility` column

i tested both the up (`npx sequelize-cli db:migrate`) and down (`npx sequelize-cli db:migrate:undo`) functionality locally and it all looks good to me: no complaints, and `events` and `current_events` both have the expected structure in each case. local deployment looks good and `web-portal` unit tests pass (would they catch db issues? there don't seem to be any unit tests in `api-server`).

### DIFF
--- a/api-server/migrations/20210709185729-add-column-for-accessibility.js
+++ b/api-server/migrations/20210709185729-add-column-for-accessibility.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+    ALTER TABLE events
+    ADD COLUMN accessibility character varying(255)[] DEFAULT (ARRAY[]::character varying[])::character varying(255)[];
+    `)
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+    ALTER TABLE events
+    DROP COLUMN accessibility;
+    `)
+  }
+};
--- a/api-server/migrations/20210709185729-add-column-for-accessibility.js
+++ b/api-server/migrations/20210709185729-add-column-for-accessibility.js
@@ -5,13 +5,108 @@ module.exports = {
     return queryInterface.sequelize.query(`
     ALTER TABLE events
     ADD COLUMN accessibility character varying(255)[] DEFAULT (ARRAY[]::character varying[])::character varying(255)[];
+
+    DROP VIEW current_events;
+
+    CREATE VIEW current_events AS
+        SELECT
+           events.id,
+           events.title,
+           events.slug,
+           events.multi_day,
+           events.image,
+           events.social_image,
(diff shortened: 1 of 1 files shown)


MatthewGidcomb:
For some reason Github won't let me formally approve, but it looks good to me.
I just have one question about the column definition, since the syntax is a bit cryptic -- is it an array of strings or a dict / map of string: string? We probably discussed this a few Sundays ago and I just don't remember.

rlmead:
ah yeah that syntax looks crazy doesn't it. the column's data type is an array of strings, with a default value of an empty array. i took the syntax from the creation of the `tags` column in the db dump i have, since we decided to structure the `accessibility` column the same way as `tags` for now. here i'm looking in the db with psql, and you can see that it describes the two fields in the same way:

```
infinite-api=# \d events
                                                    Table "public.events"
      Column       |           Type           | Collation | Nullable |                        Default                         
-------------------+--------------------------+-----------+----------+--------------------------------------------------------
 id                | uuid                     |           | not null | uuid_generate_v4()
 title             | character varying(255)   |           | not null 
...
 tags              | character varying(255)[] |           |          | ARRAY[]::character varying[]::character varying(255)[]
...
 accessibility     | character varying(255)[] |           |          | ARRAY[]::character varying[]::character varying(255)[]
```

MatthewGidcomb:
Ah. The second `::` in the default value is what made me think key: value pair, but I think I see what it's doing now.

api-server/migrations/20210709185729-add-column-for-accessibility.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+    ALTER TABLE events
+    ADD COLUMN accessibility character varying(255)[] DEFAULT (ARRAY[]::character varying[])::character varying(255)[];
```
chriswininger, line 7: I think this column could just be of type varchar(255), no?

chriswininger (approved):
One comment, I believe this is what @MatthewGidcomb was referring too in regards to the column type. I think you could probably just use varchar(255) though perhaps there's something special this gives us? At a glance I'm thinking they might be the same thing just varchar is perhaps a shortcut

rlmead:
> One comment, I believe this is what @MatthewGidcomb was referring too in regards to the column type. I think you could probably just use varchar(255) though perhaps there's something special this gives us? At a glance I'm thinking they might be the same thing just varchar is perhaps a shortcut

notes following discussion: we'd decided to make this field an array to match the `tags` field, and the syntax here matches the syntax used to create that field and set its default value.

PR #244, opened 2021-08-29
Db date restructure

rlmead:
This PR includes the initial steps toward restructuring our date data in columns rather than jsonb. The included development lays the groundwork to dismantle the dependencies on the jsonb datetime structure, but should still be compatible with production (local testing worked as expected).


api-server/migrations/20210828215333-refactor-date-storage.js
```diff
@@ -0,0 +1,146 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+      CREATE VIEW public.datetime_venue AS
+      SELECT public.uuid_generate_v4() as id,
+        id AS event_id,
+        venue_id,
+        start_time,
+        end_time,
+        optional_title
```
chriswininger: I don't think we need to spell out that it's optional in the name, that should be made clear by the type system/api swagger and such. When we define this as a table I image it will be something like

```
      CREATE TABLE if NOT EXISTS datetime_venue(
        event_id VARCHAR(255) NOT NULL
            CONSTRAINT datetime_venue_event_id REFERENCES events,
        venue_id VARCHAR(255) NOT NULL
           CONSTRAINT datetime_venue_venue_id REFERENCES venues,
       start_time timestamp with time zone NOT NULL,
       end_time timestamp with time zone NOT NULL,
       title VARCHAR(255)
      );
```

api-server/migrations/20210828215333-refactor-date-storage.js
```diff
@@ -0,0 +1,146 @@
...
+      FROM
+          (SELECT * FROM events,
+            jsonb_to_recordset(events.date_times) AS dates(start_time timestamp, end_time timestamp, optional_title character varying(255)))
```
chriswininger: that `jsonb_to_recordset` is super helpful. I can't believe I didn't know about that 👍

chriswininger (commented):
I think this all looks good, but I still think we should go ahead and make the tables. I don't think it will be a much larger step. I created caw-suggest-create-tabe off of your branch

commit 
5476198ab9dbfb9cca67e036bcf5f2a2cfb1daa0

modifies this slightly to create the tables, we can discuss

chriswininger (approved):
looks good

PR #268, opened 2022-02-03
Decouple Auth To Enable Local Testing

chriswininger:
Onboarding has been high touch. This is an open source project, accepting contributions should be simple.

To make contributing simple it should be easy to run all tests including unit and integration both locally and in ci/cd

As yet untrusted members of the public should not need secrets to run tests.

To enable the above workflow this pr decouples the admin authentication process. We rely on auth0 which is great because we don't have to store user passwords but this makes local testing hard.

These changes preserve the use of auth0 in production but offer a way to enable local test users instead for devs. Enabling test users only affects how passwords are checked and the jwt is created. All other parts of authentication remain unchanged which means that we aren't losing much coverage when it comes to integration testing.

Background Note:

One of the reasons this is possible is that we already decoupled the validation of tokens a while ago. Rather than asking auth0 on every request to validate our tokens we have been checking them against a shared pem provided by auth0. For local testing you will not need this real pem, any will do and test pem is checked into source.


api-server/src/authentication/parse-jwt.ts
```diff
@@ -1,10 +1,6 @@
-import * as fs from "fs";
 import jwt = require('jsonwebtoken');
 import {Request} from "express";
-import * as path from 'path';
-
-const PATH_TO_PEM = process.env.JWT_PEM || './keys/1nfinite.pem';
-const SECRET = fs.readFileSync(path.resolve(PATH_TO_PEM));
+import {SECRET} from "../constants";
```
MatthewGidcomb, line 3: So we don't need the PEM at all anymore?

api-server/src/main.ts
```diff
@@ -34,6 +31,10 @@ async function bootstrap() {
     await app.listen(PORT);
 
     console.info('application listening on port ', PORT);
+
+    if (AUTH_USE_TEST_USERS) {
+        console.warn(AUTH_USE_TEST_USERS_WARNING)
+    }
```
MatthewGidcomb: I wonder if we could throw an exception if `AUTH_USE_TEST_USERS` is used in combination with `ENV=production`?

web-portal/nuxt.config.js
```diff
@@ -2,6 +2,10 @@ import colors from 'vuetify/es5/util/colors'
 import 'dotenv/config'
 import { json } from 'body-parser'
 
+const API_URL = process.env.API_URL
+
+console.log('!!! ahh: ' + API_URL)
+
```
MatthewGidcomb: Did you mean to leave this? If so, can we make the message clearer?

web-portal/pages/login.vue
```diff
@@ -1,10 +1,101 @@
+<template>
+  <div class="container login-page">
+    <h1>Login</h1>
+
+    <div v-if="errorMessage !== null" class="login-page__errors">
+      {{ errorMessage }}
+    </div>
+
+    <form v-on:submit.prevent="onLoginClick">
```
MatthewGidcomb: Nitpick but `@submit.prevent=` is generally preferred over `v-on`

web-portal/pages/login.vue
```diff
@@ -1,10 +1,101 @@
...
+        }).catch((err) => {
+          console.error('error logging in:' + err)
+          this.errorMessage = err
+        })
+      }
     }
   }
 </script>
+
+<style>
```
MatthewGidcomb: We should probably scope this `<style scoped>`. It doesn't hurt but for consistency.

web-portal/store/utils/getToken.js
```diff
@@ -0,0 +1,8 @@
+const TOKEN_KEY = '_token.local'
+
+export default function getToken($auth) {
+  // according to docs this should be much simpler but this is the only way I seem to be able to get the token
+  // (https://github.com/nuxt-community/auth-module/issues/844)
+
+  return $auth.$storage.getState(TOKEN_KEY)
+}
```
MatthewGidcomb: `utils` probably needs to be a top-level directory, and not in `store/`. Per the [docs](https://nuxtjs.org/docs/directory-structure/store), every `.js` file in `store/` is treated as a Vuex store module.

web-portal/store/utils/getToken.js
```diff
@@ -0,0 +1,8 @@
+const TOKEN_KEY = '_token.local'
+
+export default function getToken($auth) {
+  // according to docs this should be much simpler but this is the only way I seem to be able to get the token
+  // (https://github.com/nuxt-community/auth-module/issues/844)
+
+  return $auth.$storage.getState(TOKEN_KEY)
```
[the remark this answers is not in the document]
chriswininger: ignore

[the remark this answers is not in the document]
chriswininger: @sonatype-lift ignore

api-server/_constants.js
```diff
@@ -0,0 +1,81 @@
...
+      './test-keys/1nfinite_testing_rsa256.rsa' :
+      './keys/1nfinite.pem';
+}
+
+function isTrueFromString(value) {
+  return (value || '').toLowerCase() === 'true';
+}
+
+function isNullOrUndefined(value) {
+  return value === undefined || value === null
```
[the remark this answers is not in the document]
chriswininger: @sonabot-lift ignore -- you semicolon loving monster

api-server/src/authentication/parse-jwt.ts
chriswininger: we still need the it's just being imported and potentially defaulted inside of constants now so that we don't have to repeat the defaulting logic

web-portal/nuxt.config.js
chriswininger: whoops, haha I was about to say: "At least there are no swear words this time", then I found this one

```
console.log('!!! what the shit')
console.log(VenueModel)
console.log("_________")
```

chriswininger: cleaning it up now :-)

web-portal/store/utils/getToken.js
chriswininger: :-( OK, good catch. I'll move it. I'm not a huge fan of how nuxt makes us spread out our implementation details. NextJs has the same problem. It seems like they could give you an annotation to let you exclude some things

web-portal/helpers/getToken.js
```diff
@@ -0,0 +1,8 @@
+const TOKEN_KEY = '_token.local'
+
+export default function getToken($auth) {
+  // according to docs this should be much simpler but this is the only way I seem to be able to get the token
+  // (https://github.com/nuxt-community/auth-module/issues/844)
+
+  return $auth.$storage.getState(TOKEN_KEY)
```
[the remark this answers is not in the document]
chriswininger: @sonatype-lift ignore

PR #271, opened 2022-02-13
Fix datetime updates

rlmead:


api-server/src/events/events.service.ts
```diff
@@ -47,14 +47,47 @@ export class EventsService {
...
+
+            const updatedEvent = await this.eventModel.update(values, updateQueryOptions)
+
+            await this.updateDatetimeVenueEntries(
+                id,
+                values.venue_id,
+                values.date_times,
+                transactionHost);
+
+            return updatedEvent;
```
MatthewGidcomb, line 68: This probably isn't important because I don't think the frontend uses this for anything, but I wonder if `updatedEvent` will have the old datetimes?

MatthewGidcomb (approved):

api-server/src/events/events.service.ts
chriswininger: currently the frontend may not, but I think it's a good pattern to always return the updated state. It allows the client to know about any autogenerated fields such ids or timestamps without having to refetch.

If we wanted to be complete in our use of the pattern we should probably create a return type that allows us to mix back in the final state of the DateTimes.

It could be something like

```
EventUpdateResponse {
   eventModel: EventModel,
   date_times: DatetimeVenueModel [];
}
```

In any event, like Mathew says, the frontend doesn't currently do anything with this; so, not that important. I'm mostly pointing it out to explain the general pattern I've been trying to follow with endpoints.

Feel free to tweak it some more or not. There will always be more to refine :-)

chriswininger (approved):
looks good to me, left one minor comment alongside Mathews

chriswininger:
If we weren't updating the DateTimeVenueTable before, how were we updating the start_end times in the recent past following the refactor?

PR #277, opened 2022-03-27
Admins Can Flag Events as Having a Problem

chriswininger:
This gives admins the ability to check a box indicating that an event has a problem that is being investigated.

To support this a new table is created for event admin metadata. Currently it's just used for this flag but could be expanded on in the future.

We can easily exclude this table from results for non-admins.

In most cases we are able to return the flag as part of an event through the inclusion query param. This was not possible with current_events due to implementation details. This should be fixed soon with other in progress work.

Video Demo: https://youtu.be/P1B2MSbWs7U

### DIFF
--- a/api-server/src/events/events.authenticated.controller.ts
+++ b/api-server/src/events/events.authenticated.controller.ts
@@ -26,7 +26,6 @@ export default class EventsAuthenticatedController {
     // this probably shouldn't be needed long term, we should just fetch this with the event
     @Get('/admin-metadata')
     @ApiOperation({ summary: 'Get all admin metadata for all events' })
-    @ApiImplicitParam({ name: 'id', type: String })
     getEventAdminMetadata(): Promise<EventAdminMetadataListResponse> {
         return this.eventsService.getAllEventMetaData()
             .then((eventAdminMetadata) =>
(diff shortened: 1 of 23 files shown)


web-portal/store/event-admin-metadata.js
```diff
@@ -0,0 +1,40 @@
...
+    return ApiService.put(`/authenticated/events/${eventId}/admin-metadata`, { isProblem }, idToken)
+      .then((response) => {
+        context.commit('admin/UPDATE_ADMIN_METADATA', { eventId, newMetaDateEntry: response.data.eventAdminMetadata }, { root: true })
+        context.commit('GET_UPSERT_EVENT_METADATA_FETCH_SUCCESS', response.data.eventAdminMetadata)
+      })
+      .catch((error) => {
+        context.commit('GET_UPSERT_EVENT_METADATA_FETCH_FAIL', error)
+      })
+  }
+}
```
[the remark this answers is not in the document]
chriswininger: @sonatype-lift ignore

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -21,6 +23,16 @@ export default class EventsAuthenticatedController {
 
     constructor(private readonly eventsService: EventsService){}
 
+    // this probably shouldn't be needed long term, we should just fetch this with the event
+    @Get('/admin-metadata')
+    @ApiOperation({ summary: 'Get all admin metadata for all events' })
+    @ApiImplicitParam({ name: 'id', type: String })
```
MatthewGidcomb: Is this annotation supposed to be here? Doesn't look like `id` is used.

MatthewGidcomb (approved):

api-server/src/events/events.authenticated.controller.ts
chriswininger: nope, cpy/paste fail, good catch

PR #280, opened 2022-04-03
Drop current events

rlmead:


api-server/migrations/20220320153148-add-fields-to-current_date_times-view.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+    CREATE OR REPLACE VIEW current_date_times_jsonb AS
```
chriswininger: Do we still need this first migration? Aren't we dropping this in the next migration?

api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
```diff
@@ -1,11 +1,9 @@
 import {Request} from "express";
-import {CurrentEvent} from "../../current-events/models/current-event.model";
 import isAdminUser from "../is-admin-user";
-import {VenueModel} from "../../venues/models/venue.model";
 import EventDTO from "../../events/dto/eventDTO";
 
-type GenericEvent = EventDTO | CurrentEvent
-type GenericEventList = EventDTO[] | CurrentEvent[]
```
chriswininger, line 8: I like seeing this disappear, very nice

image

api-server/src/events/events.controller.ts
```diff
@@ -29,17 +31,54 @@ export class EventsController {
...
+    ): Promise<EventsResponse> {
+        if (typeof embed === 'string') {
+            embed = [embed, 'DATE_TIME']
+        } else {
+            embed.push('DATE_TIME')
+        }
+
+        const findOptions = {
+            ...getOptionsForEventsServiceFromEmbedsQueryParam(embed),
+            where: {
```
chriswininger, line 55: nice, thinks for getting this smoothed out.

Do you think this is best left here or shifted to a service method like getCurrentEvents?

api-server/src/events/events.controller.ts
```diff
@@ -29,17 +31,54 @@ export class EventsController {
...
     @Get('verified')
-    @ApiOperation({summary: 'Get current events that have been verified (for public consumptions)'})
+    @ApiOperation({ summary: 'Get all events that have been verified (for public consumptions)' })
     @ApiResponse({
         status: 200,
         description: 'verified events',
         type: EventsResponse
     })
-    getAllCurrentVerified(
+    getAllVerified(
```
chriswininger, line 81: great to see this confusing mess get cleaned up :+1:

widget/src/apiService.js
```diff
@@ -1,7 +1,7 @@
 export default class APIService {
 
     static getEvents(context, callback) {
-        APIService.get(`${context.getApiUrl()}/current-events/verified?embed=Venue`, function(err, response) {
+        APIService.get(`${context.getApiUrl()}/events/current-verified?embed=Venue`, function(err, response) {
```
chriswininger, line 4: good catch, you even updated the widget :-) I'm not sure if we use this for anything right now, but who knows maybe someday

chriswininger (approved):
This looks great!!!

It looks like there may be an nu-needed migration you forgot to delete, but doubt it hurts anything.

I did a side by side of local with prod db snapshot vs prod. I noticed one minor difference, but I think it's an improvement at best and nothing to worry at about at least.

When two events happen at the same time the order was different, but in this version they appear to sort alphabetical in that scenario whereas before  I suspect it was just based on ID or something; so, I'd say that's better.

Prod is on left; Local is on right;
Screenshot from 2022-04-09 10-26-59

api-server/migrations/20220320153148-add-fields-to-current_date_times-view.js
rlmead: good call, deleted 👍🏼

api-server/src/events/events.controller.ts
rlmead: yeah i agree there's room for improvement here in future refactors 👍🏼

PR #285, opened 2022-05-05
Update EventCard to reflect new datetime-venue structure

MatthewGidcomb:
Venue is now associated with datetime instead of event; the card component needed to be updated to reflect this.

May need to discuss this: in particular, do we want to tweak the way the venues are exposed on the event, when embedded?


chriswininger (approved):
looks good

PR #290, opened 2022-06-12
Publish to docker hub

chriswininger:
this pr is no where near ready but I need to make it a pr to test as per https://stackoverflow.com/questions/63362126/github-actions-how-to-run-a-workflow-created-on-a-non-master-branch-from-the-wo

### DIFF
--- a/.github/workflows/publish-to-docker-hub.yml
+++ b/.github/workflows/publish-to-docker-hub.yml
@@ -1,0 +1,33 @@
+on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+        type: choice
+        options:
+          - info
+          - warning
+          - debug
+      tags:
+        description: 'Test scenario tags'
+        required: false
+        type: boolean
+      environment:
+        description: 'Environment to run tests against'
+        type: environment
+        required: true
+
+jobs:
+  log-the-inputs:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Log level: $LEVEL"
+          echo "Tags: $TAGS"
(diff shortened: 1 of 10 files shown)


api-server/docker-publish.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
+tag="${1:=local}"
+container_name="infinite-api-server"
+dockerhub_org="chriswininger"
```
MatthewGidcomb, line 7: Could read the org name from an env variable, and then put it in a Github secret; then if/when we create an Infinite Dockerhub org we just have to change it in one place.

.github/workflows/publish-to-docker-hub.yml
```diff
@@ -0,0 +1,50 @@
+on:
+  workflow_dispatch:
+    environment: docker-publication
```
MatthewGidcomb, line 3: What does using an environment here give us?

api-server/docker-publish.sh
chriswininger: I don't think it really needs to be a secret. Nothing secret about it and this keeps the script flexible for local host, at least that was my thinking

.github/workflows/publish-to-docker-hub.yml
chriswininger: well the TEST_SECRET one I need to delete but the COMMIT_HASH one just makes it availabe in the script for

`./docker-build.sh "$COMMIT_HASH"`

I think I could also just do:

`./docker-buid.sh {inputs.commit}`

I'm honestly not sure which is better but the tutorial I followed mapped the inputs and secrets to env for some reason, so really just copy/paste

MatthewGidcomb (approved):

PR #299, opened 2022-10-09
Use timezone info in web portal

rlmead:


web-portal/test/EventCard.spec.js
```diff
@@ -86,21 +86,9 @@ describe('Card component', () => {
...
-    wrapper.setProps({
-      calendar_event: Object.assign(getEvent(), {
-        date_times: [
-          { start_time: '2020-06-01T14:00:00.000Z', end_time: '2020-06-01T15:00:00.000Z' }
-        ]
-      })
-    })
-    expect(wrapper.html()).toContain('June 1st')
-    expect(wrapper.html()).toContain('10:00am - 11:00am')
+    expect(wrapper.html()).toContain('10:00am - 11:00am EDT')
```
rlmead, line 96: mysteriously, this test wasn't failing on my local machine 🤔 i didn't find this problem until this test ran on github. no idea why that is. it's an obvious one though :relaxed:

also, there used to be separate tests for datetimes stored with and without timezones. since we expect all datetimes to be stored as timezone utc with timezone specified separately, i consolidated the two tests into one updated test.

web-portal/components/DateTimePicker.vue
```diff
@@ -77,6 +77,11 @@
                 <option value="am">AM</option>
                 <option value="pm">PM</option>
               </select>
+              <select ref="eventTimezone" name="event_timezone" v-model="event_timezone">
+                <option v-for="(tz) in this.$config.TIMEZONE_OPTIONS.split(',')" :key="tz">
```
MatthewGidcomb: Small nitpick but the `this` isn't technically necessary inside the `template` -- you should be able to refer to it as just `$config.TIMEZONE_OPTIONS...`.

web-portal/components/EventView.vue
```diff
@@ -20,12 +20,10 @@
...
         <div v-if="!isOnlineResource">
           <div v-for="(date_time, index) in event.date_times" :key="index" class="date-time-container">
-            <!-- TODO: do we need to consider timezones? If so, this might be useful: -->
-            <!-- https://stackoverflow.com/a/57022505 -->
             <!-- TODO: should we consider Luxon instead of Moment? -->
-            <em>{{ date_time.start_time | moment('dddd, MMMM Do') }}</em>
+            <em>{{ adjustDatetimeForTimezone(date_time.start_time, date_time.timezone).format('dddd, MMMM Do') }}</em>
             <br>
-            <em>{{ date_time.start_time | moment('h:mma') }} - {{ date_time.end_time | moment('h:mma') }}</em>
+            <em>{{ adjustDatetimeForTimezone(date_time.start_time, date_time.timezone).format('h:mma -') }} {{ adjustDatetimeForTimezone(date_time.end_time, date_time.timezone).format('h:mma z') }}</em>
```
MatthewGidcomb: We should consider refactoring this to use computed properties, similar to the `when_date` and `when_time` in `EventCard`.

web-portal/components/EventView.vue
```diff
@@ -212,6 +223,9 @@
         console.error('Action:', e.action)
         console.error('Trigger:', e.trigger)
         window.alert('We were unable to copy URL. Here it is for reference:\n' + this.event.bitly_link)
+      },
+      adjustDatetimeForTimezone(datetime, timezone) {
+        return momenttz(datetime).tz(timezone)
```
MatthewGidcomb: I don't see any other references to this in the PR. If it ended up not being used let's remove it.

MatthewGidcomb (approved):

web-portal/components/EventView.vue
rlmead: good catch! done, thanks ✅

api-server/migrations/20220814155014-change_default_timezone.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+      ALTER TABLE datetime_venue
+      ALTER COLUMN timezone
+      SET DEFAULT 'US/Eastern';
+
+      UPDATE datetime_venue
+      SET timezone = 'US/Eastern'
+      WHERE timezone = 'America/New_York';
```
chriswininger, line 12: nice cleanup

api-server/migrations/20220814155014-change_default_timezone.js
```diff
@@ -0,0 +1,27 @@
...
+
+  down: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+      ALTER TABLE datetime_venue
+      ALTER COLUMN timezone
+      SET DEFAULT 'America/New_York';
+
+      UPDATE datetime_venue
+      SET timezone = 'America/New_York'
+      WHERE timezone = 'US/Eastern';
```
chriswininger, line 24: does this really work in reverse? What if someone had entered the timezone as US/Eastern?

web-portal/.env.sample
```diff
@@ -19,3 +19,7 @@ API_URL=http://localhost:3003/v1
 
 # Slack Contact Form
 SLACK_WEBHOOK_CONTACT=
+
+# Timezone handling
+TIMEZONE_DEFAULT=
```
chriswininger, line 24: thx for updating

web-portal/README.md
```diff
@@ -46,6 +46,29 @@ This application communicates with an [API server](https://github.com/infinite-i
  cp .env.sample .env    # main env file
  ```
 
+* The options you choose for your time zone variables should meet three criteria:
```
chriswininger, line 49: +1 for documentation :-)

chriswininger (approved):
+1

api-server/migrations/20220814155014-change_default_timezone.js
rlmead: @chriswininger i don't think i see the problem you see here. in this down-migration, all timezones with a value of `US/Eastern` get rolled back to `America/New_York`, which is correct. this migration just marks a change in the set of timezones and the default that we're using for our instance of the app - to date, everything has been getting the default `America/New_York`, but we decided to change the set we're using. no humans have been able to select a timezone for their datetime so far (and even after they can, they'll only be able to choose from the subset we give them).

or are you asking this from the perspective of "`US/Eastern` and `America/New_York` are not identical"? if that's what you're asking, then the same question goes for the up-migration. this is kind of a can of worms, and we landed on `US/Eastern` for a few reasons, but we can revisit that conversation in this weekend's meeting if you want.

PR #309, opened 2022-11-13
Timezone followup

rlmead:
Addresses two timezone issues:
- display timezone on admin events list
- use specified timezone in timezone-selector dropdown when updating a datetime

### DIFF
--- a/web-portal/components/DateTimePicker.vue
+++ b/web-portal/components/DateTimePicker.vue
@@ -166,10 +166,10 @@
 
 <script>
 
-  import moment from 'moment-timezone'
+  import momenttz from 'moment-timezone'
 
   // this is how the date/time is stored in data and sent to the server
-  const dateTimeStorageFormat = moment.ISO_8601
+  const dateTimeStorageFormat = momenttz.ISO_8601
 
   // this format is used for parsing date/times extracted from the picker before storing them
   const dateTimePickerFormat = 'YYYY-MM-DD hh:mm:a zz'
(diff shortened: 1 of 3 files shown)


web-portal/components/DateTimePicker.vue
```diff
@@ -263,14 +263,15 @@
...
-        this.end_ampm = moment(time_segment.end_time).format('a')
+        this.picker = momenttz(time_segment.start_time).tz(time_segment.timezone).format('YYYY-MM-DD')
+        this.start_hour = momenttz(time_segment.start_time).tz(time_segment.timezone).format('hh')
+        this.start_minute = momenttz(time_segment.start_time).tz(time_segment.timezone).format('mm')
+        this.start_ampm = momenttz(time_segment.start_time).tz(time_segment.timezone).format('a')
+
+        this.end_hour = momenttz(time_segment.end_time).tz(time_segment.timezone).format('hh')
+        this.end_minute = momenttz(time_segment.end_time).tz(time_segment.timezone).format('mm')
+        this.end_ampm = momenttz(time_segment.end_time).tz(time_segment.timezone).format('a')
+        this.event_timezone = time_segment.timezone
```
MatthewGidcomb: I wonder if this would be easier to read if the `momenttz(...).tz(...)` part was saved to a variable, and each subsequent assignment was just e.g. `time.format(...)`?

Might also improve performance, although we don't have any particular reason to believe it's a problem.

MatthewGidcomb (approved):

web-portal/components/DateTimePicker.vue
rlmead: great idea, will do 💡

web-portal/components/DateTimePicker.vue
```diff
@@ -263,14 +263,16 @@
...
-        this.picker = moment(time_segment.start_time).format('YYYY-MM-DD')
-        this.start_hour = moment(time_segment.start_time).format('hh')
-        this.start_minute = moment(time_segment.start_time).format('mm')
-        this.start_ampm = moment(time_segment.start_time).format('a')
-
-        this.end_hour = moment(time_segment.end_time).format('hh')
-        this.end_minute = moment(time_segment.end_time).format('mm')
-        this.end_ampm = moment(time_segment.end_time).format('a')
+        this.event_timezone = time_segment.timezone
+        const start_time = momenttz(time_segment.start_time).tz(this.event_timezone)
```
chriswininger, line 268: small style suggestion, this might be easier to read with some new lines. Exactly where they go is subjective. I  usually try to split on related units. In this case I might stick new lines above the start_time and end_time `const` declarations

chriswininger (approved):
+1 nice

MatthewGidcomb (approved):

PR #311, opened 2022-11-20
Venue Edit Page Enhancements (#310)

chriswininger:
Added the ability to update values directly on existing venues from the venue admin editor page. This will be useful for fixing typos

Added the ability to create a new venue from existing entries while soft deleting the old venue. This will be useful for when venues move, where we may want to keep old events associated with the past address while pointing new events to the new address.

Improved the styling to make it more clear that fields are editable

[Demo Video](https://youtu.be/Qi57YJfPl0w)


chriswininger:
Release Development

rlmead (approved):

PR #312, opened 2022-11-20
add ansible scripts to infinite repo

chriswininger:
Moves the ansible scripts I used to setup up our servers into the main repo


rlmead (approved):

ansible/hosts
```diff
@@ -0,0 +1,20 @@
...
+
+[infinite-prod]
+23.100.45.102
+
+[infinite-prod:vars]
+ansible_port=22
+ansible_user=infinite
+
+[infinite-local]
+192.168.1.43
```
MatthewGidcomb, line 16: Do these IP addresses need to be hardcoded?

MatthewGidcomb (approved):

ansible/hosts
chriswininger: yeah I think it's best to go with the ips for. I went back and fourth on this a bit, but I had an accident with DNS pointing the wrong box once that made we want to stick to ips. We can revisit as things mature

PR #314, opened 2022-11-27
Add GPS Coordinate Fields

chriswininger:
- Add lat/long/altitude fields to the venues table
- Create endpoint that can deduce lat/long/altitude from google maps link
- Update admin form to allow entry of lat/long/altitude
- Update admin form to allow resolution of lat/long/altitude from google maps link


rlmead (approved):

api-server/src/venues/dto/create-update-venue-request.ts
```diff
@@ -13,6 +14,26 @@ export class CreateVenueRequest {
...
+
+    @Column
+    @ApiProperty({example: 17.1 })
+    @IsOptional()
+    gps_alt?: number
+
+    @Column
+    @ApiProperty({ example: false })
+    @IsOptional()
+    is_soft_deleted: boolean;
```
MatthewGidcomb: Is there a reason the create request would include the soft delete flag?

web-portal/components/admin-venue-edit/VenueCard.vue
```diff
@@ -153,6 +199,27 @@
 
         this.$store.dispatch(DELETE_VENUE, { id, idToken })
       },
+      onFindUsingMapLinksClick: function () {
+        this.$data.fetchingGpsCoordinates = true
+        this.$data.fetchingGpsCoordinatesError = null
```
MatthewGidcomb: Nitpick here but there's no reason for the `$data` part -- should be just `this.fetchingGpsCoordinates = true` etc

api-server/src/venues/venues.controller.ts
```diff
@@ -69,6 +72,21 @@ export class VenuesController {
...
+        description: 'gps from google maps links',
+        type: GetGPSCoordinatesResponse
+    })
+    getGpsFromGoogleMapsLink(@Body() req: GetGPSCoordinatesRequest): Promise<GetGPSCoordinatesResponse> {
+        const { googleMapsLink } = req;
+
+        return this.gpsService.getCoordinatesFromGoogleMapLink(googleMapsLink).then((gpsCoordinates) => new GetGPSCoordinatesResponse({
+            gpsCoordinates
+        }))
+    }
```
MatthewGidcomb, line 89: Maybe I'm missing something, but why is this a `post`? Doesn't appear to create or change anything server-side.

Although I wonder if it would make sense (and if Nest even supports this...) to support both `get` and `put` for this endpoint; `get` just queries and returns the coordinates, `put` queries and updates the venue with the coordinates without another round-trip to the client.

api-server/src/venues/dto/create-update-venue-request.ts
chriswininger: Hmm probably not ideal, should just get defaulted to false on create. I wonder if that was copy/paste mistake on my part or if this dto is overloaded. I'll poke at this  a bit

api-server/src/venues/venues.controller.ts
chriswininger: I went this route mouse because I was thining queryString felt weird given the url is required, but I  also didn't want to mess with making sure a url was properly serialized as a path param. PostBody felt like the best place to put it, but you're right probably makes more sense as a get

web-portal/components/admin-venue-edit/VenueCard.vue
chriswininger: ok cool good to know, will update

api-server/src/venues/dto/create-update-venue-request.ts
chriswininger: yeah not sure why I added, that, fixing now

MatthewGidcomb (approved):

PR #333, opened 2022-12-07
Feature: add calendar page

pain-x01:
Here is an implementation of the 'Fullcalendar' https://www.npmjs.com/package/@fullcalendar/vue. After some investigation, it seemed to best solution, which corresponds to the listed requirements. V-Calendar is more suitable as a date picker component and requires a lot of customization, tui.calendar doesn't look like what we want at all. Also, I can recommend Vuetify calendar component version 2, but the current version used on site ("vuetify": "1.5.24") doesn't support events overlapping (like google calendar). Please check my merge-request

### DIFF
--- a/web-portal/package.json
+++ b/web-portal/package.json
@@ -17,6 +17,12 @@
     "generate": "nuxt generate"
   },
   "dependencies": {
+    "@fullcalendar/core": "^5.11.3",
+    "@fullcalendar/daygrid": "^5.11.3",
+    "@fullcalendar/interaction": "^5.11.3",
+    "@fullcalendar/moment": "^5.11.3",
+    "@fullcalendar/timegrid": "^5.11.3",
+    "@fullcalendar/vue": "^5.11.3",
     "@nuxtjs/auth": "4.9.1",
     "@nuxtjs/axios": "^5.12.0",
     "@nuxtjs/google-gtag": "^1.0.4",
--- a/web-portal/package.json
+++ b/web-portal/package.json
@@ -20,7 +20,6 @@
     "@fullcalendar/core": "^5.11.3",
     "@fullcalendar/daygrid": "^5.11.3",
     "@fullcalendar/interaction": "^5.11.3",
-    "@fullcalendar/moment": "^5.11.3",
     "@fullcalendar/timegrid": "^5.11.3",
     "@fullcalendar/vue": "^5.11.3",
     "@nuxtjs/auth": "4.9.1",
--- a/web-portal/pages/calendar.vue
+++ b/web-portal/pages/calendar.vue
@@ -42,16 +42,16 @@
       }
     },
     async fetch({ store }) {
-      await store.dispatch('LoadAllCalendarEventData')
+      await store.dispatch('LoadAllLocalEventData')
     },
     computed: {
-      ...mapGetters({ getAllCalendarEvents: 'GetAllCalendarEvents' })
+      ...mapGetters({ getAllLocalEvents: 'GetAllLocalEvents' })
     },
     mounted() {
       this.setUpdateCalendarEvents()
 
       this.timeInterval = setInterval(() => {
-        this.$store.dispatch('LoadAllCalendarEventData').then(() => {
+        this.$store.dispatch('LoadAllLocalEventData').then(() => {
           this.setUpdateCalendarEvents()
         })
       }, 5 * 60 * 1000)
@@ -60,7 +60,7 @@
       setUpdateCalendarEvents() {
         const parsedEvents = []
 
-        this.getAllCalendarEvents.forEach((event) => {
+        this.getAllLocalEvents.forEach((event) => {
           event.date_times.forEach((eventDate) => {
             parsedEvents.push({
               title: event.title,
--- a/web-portal/pages/calendar.vue
+++ b/web-portal/pages/calendar.vue
@@ -22,7 +22,7 @@
     data: function () {
       return {
         timeInterval: null,
-        calendarOptions: {
+        initialCalendarOptions: {
           plugins: [dayGridPlugin, timeGridPlugin, interactionPlugin],
           initialView: 'dayGridMonth',
           nowIndicator: true,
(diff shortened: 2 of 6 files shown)


web-portal/package.json
```diff
@@ -17,6 +17,12 @@
     "generate": "nuxt generate"
   },
   "dependencies": {
+    "@fullcalendar/core": "^5.11.3",
+    "@fullcalendar/daygrid": "^5.11.3",
+    "@fullcalendar/interaction": "^5.11.3",
+    "@fullcalendar/moment": "^5.11.3",
```
MatthewGidcomb: So I gather we need the `interaction` plugin for clicking on events, but do we actually need the `moment` plugin?

web-portal/pages/calendar.vue
```diff
@@ -0,0 +1,109 @@
...
+              end: eventDate.end_time,
+              groupId: event.id,
+              editable: false,
+              originId: event.id
+            })
+          })
+        })
+
+        this.calendarOptions.events = parsedEvents
+      },
```
MatthewGidcomb: Thoughts on refactoring this as a computed property? It would only be re-evaluated when the `getAllCalendarEvents` getter updated.
Would simplify the interval, too -- just need to fire off the store action and let Vue do the rest.

web-portal/pages/calendar.vue
```diff
@@ -0,0 +1,109 @@
...
+}
+.calendar-page .fc-daygrid-event {
+  color: #5c5c5c;
+}
+.calendar-page .fc-col-header-cell-cushion {
+  color: #000;
+}
+.calendar-page .fc-daygrid-day-number {
+  color: #000;
+}
```
MatthewGidcomb: Tiny nitpick but we usually indent the `style` block one level (two spaces)

web-portal/store/index.js
```diff
@@ -113,7 +119,19 @@ export const actions = {
...
+    return this.$apiService.get(CURRENT_EVENTS_VERIFIED_PATH)
+      .then((_response) => {
+        context.commit('UPDATE_CALENDAR_EVENTS', _response.data.events)
+        context.commit('SET_LOADING_STATUS', false)
+      })
+      .catch((error) => {
+        console.error(error)
+        context.commit('ui/SHOW_NOTIFICATIONS', { open: true, message: 'Hrrmm... unable to get event data. Please contact us and we will figure out what went wrong.' }, { root: true })
+      })
+  },
```
MatthewGidcomb: Do we need this second code path for getting the calendar events? Seems like the `LoadAllLocalEventData` path would pull in everything we need -- current + verified events.

web-portal/pages/calendar.vue
```diff
@@ -0,0 +1,109 @@
...
+            left: 'dayGridMonth,dayGridWeek,timeGridDay',
+            center: 'title',
+            right: 'prev,next'
+          },
+          titleFormat: { year: 'numeric', month: 'short', day: 'numeric' },
+          dateClick: this.handleDateClick,
+          eventClick: this.handleEventClick,
+          eventBackgroundColor: '#b7b09c',
+          eventBorderColor: '#fff',
+          eventTextColor: '#fff'
```
MatthewGidcomb: Good call on using this accent color for the event backgrounds. It seems like it's only used for events that span days (but I'm guessing that's a FullCalendar behavior).

web-portal/package.json
pain-x01: @MatthewGidcomb yes, it is an extra plugin, not necessary to use it now. Removed it 25b402e40d42d2098a0f02b28bfa64ffcbc8a962

web-portal/pages/calendar.vue
pain-x01: @MatthewGidcomb got it 7e3e8905423b7014a0ca0f147a4b6a4d61d6244b

web-portal/pages/calendar.vue
pain-x01: @MatthewGidcomb Yes, indeed. I believe we can add almost any style for each element, but need to clarify the requirements to design and expected behavior for example on hover, etc.

web-portal/store/index.js
pain-x01: I made it for future optimization for calendar queries or additional params for the calendar data (like ?embed=Venue used for local events) but no problem with it, I removed extra store props 9a76f5848a479dc3e571d1f062e91c37e6d57c3a

web-portal/pages/calendar.vue
pain-x01: @MatthewGidcomb good point! a3c6f98055a60de0dfb28d94cad1864740a024ed

chriswininger:
Currently this calendar is populated with current events. At some point in the future I would like to extend it to show past events. We will not be able to fetch all events for all time because that will be too much data to fetch up front. I plan to create an end point that fetches data in a range. Will it be easy to update this to fetch new data when navigating to the previous or next month?

pain-x01:
> Currently this calendar is populated with current events. At some point in the future I would like to extend it to show past events. We will not be able to fetch all events for all time because that will be too much data to fetch up front. I plan to create an end point that fetches data in a range. Will it be easy to update this to fetch new data when navigating to the previous or next month?

@chriswininger I'm sure it won't take much time and effort. And it will be more work with the page component and vuex than with the calendar

Digi-D:
Thank you and got a couple of questions/requestes regarding calendar functionality:

1. Would it be possible to lock the interface so (at least short-term) we are not able scroll back into past? Currently it is not populated and might be confusing to the users, especially in week/day view.

2.  For mobile view, can we make the current day default and stretch it down the length of the view? This is not going to be the primary use case as real planning will be done on desktop. We just need it to look "good enough" - the calendar might get shared to new users via email and some (sigh...) will open it on the phone.
Screen Shot 2022-12-13 at 11 21 14 AM

3. Currently titles are cut off if they are too long in Month/Week views. Can we make them wrap to a new line in desktop view? 
Screen Shot 2022-12-13 at 11 17 53 AM

pain-x01:
@Digi-D Hi! I made changes accordingly to your message, I hope it is what you expect, please check my last commit

Digi-D:
> 

@pain-x01 One more adjustment related to number 2 - need title to appear across the top of the screen in mobile views 
Screen Shot 2022-12-16 at 11 37 10 AM Let me know if there are any problems that this might cause. I didn't look too close at the fullcal code and was just tweaking styles in browser to create this view.

The core group will meet Sunday to finalize and merge this branch. Thank you!

pain-x01:
@Digi-D Done :) cbec7a4234205b5f9ed387e964da58db5c7028af

rlmead (approved):

PR #341, opened 2023-01-04
Add "sold out" control tag

### LINKED ISSUE #324: Add "sold out" control tag
Similar to the existing `postponed` and `cancelled` -- when tag is applied, "[SOLD OUT]" is prepended to the event title when displayed on event cards and the event view. Opacity change probably not necessary but we can experiment.

Precedence: `cancelled` > `sold out` > `postponed`
(e.g. if an event has more than one of these tags applied, `cancelled` will be displayed before `sold-out`, etc.

While we're doing this, it might be worth it to migrate the `cancelled` and `postponed` tags to be colon-namespaced like the `mode:` and `category:` ones.

MatthewGidcomb:
Similar to postponed / cancelled, but colon-namespaced like the newer mode and category control tags

references #324

### DIFF
--- a/api-server/migrations/20230104022346-namespace-condition-tags.js
+++ b/api-server/migrations/20230104022346-namespace-condition-tags.js
@@ -1,0 +1,33 @@
+'use strict';
+
+/**
+ * Special behavior was added to the web portal for events tagged with
+ * "postponed" and "cancelled" in 2020.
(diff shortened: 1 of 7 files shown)


chriswininger (approved):
looks good to me

rlmead (approved):

PR #351, opened 2023-01-29
Jswank spelunk

chriswininger:


api-server/Dockerfile
```diff
@@ -1,15 +1,16 @@
-FROM node:gallium
+FROM node:19-alpine
```
chriswininger, line 1: a little nervous about upgrading and changing bases for this in conjunction with other changes. I'd  want to test this on it's own first and make sure everything is good given this will affect production

api-server/Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+  bash just
 
-RUN npm ci
-RUN npm run build
+COPY package.json package-lock.json ./
+COPY local.justfile ./justfile
+RUN just install-deps
 
-EXPOSE 3003
+COPY . .
```
chriswininger, line 12: is this replaceing ADD ./ /api-server? Why copy vs ADD? Is their a performance difference between these two operations?

api-server/Dockerfile
```diff
@@ -1,15 +1,16 @@
...
 
-RUN chmod +x ./docker-scripts/start.sh
-RUN chmod +x ./docker-scripts/wait-for-it.sh
-RUN npm install -g npm@8.7.0
+RUN apk -U --no-cache add \
+  bash just
 
-RUN npm ci
-RUN npm run build
+COPY package.json package-lock.json ./
```
chriswininger, line 8: won't this be copied with COPY . . ?

api-server/README.md
```diff
@@ -2,30 +2,21 @@
...
-### Setup
-
-Set up the env file and add *1nfinite.pem* file to keys directory. Sample files are provided with a `.sample` suffix.
-A core team member can send you the dev secrets via the Infinite Industries Slack.
-
- ```bash
- cp .env.sample .env
- cp keys/1nfinite.pem 1nfinite.pem    
- ```
+* [Installation](#installation)
```
chriswininger, line 5: I like the table of contents approach here

api-server/README.md
```diff
@@ -78,6 +69,121 @@ properly configured, with the appropriate .env file and pem file in its home dir
...
+$ cp keys/1nfinite.pem 1nfinite.pem
+$ just build start
+```
+
+Notes:
+* A `.env` file is used to set values for many variables. Multiple tools in the
+  toolchain- `just`, `docker-compose`, and `npm` and `node` (via
+  [dotenv](https://www.npmjs.com/package/dotenv)) all use this file. Different
+  values are used for different enviroments.  The `.env.sample` includes variables
+  appropriate for local development.
```
chriswininger, line 94: Actually for local development the idea is you shouldn't need anything here. That said defaults probably assuming local development on the host. For running inside a docker-compose stack some values need to be overridden but I would do that via container host rather than creating a .env in the container. Mostly the .env.sample exists to document the values you can set

api-server/README.md
```diff
@@ -78,6 +69,121 @@ properly configured, with the appropriate .env file and pem file in its home dir
...
+* The key `1nfinite.pem` is used for ??
+* `just` is used as a task runner: the *build* recipe create a docker image and
+  the *start* recipe starts it and a postgres database as docker containers.
+
+Other `just` recipes can be displayed by invoking it without arguments:
+
+```console
+$ just
+Available recipes:
```
chriswininger, line 103: This all looks pretty cool. I see `start a local docker-based environment via docker-compose`. Does that include the frontend?

api-server/README.md
```diff
@@ -78,6 +69,121 @@ properly configured, with the appropriate .env file and pem file in its home dir
...
+```
+
+Overriding the image name specified in `.env`:
+```console
+$ IMAGE_NAME=chriswininger/infinite-api-server just publish
+```
+
+## Development Environment Setup [OLD]
+
+### Dependencies
```
chriswininger, line 168: This section is missing from the table of contents fyi

api-server/TODO.md
```diff
@@ -0,0 +1,33 @@
...
+* Wrapping docker-compose w/ Makefile example: https://medium.com/freestoneinfotech/simplifying-docker-compose-operations-using-makefile-26d451456d63
+* Example docker-compose.yml files:
+  * sonatype/ossrsh-proxy has a reasonable starting docker-compose.yml file (builds image)
+  * links.ng/docker-compose.yml
+
+* Perhaps use `just` rather than `make`? 
+  * https://just.systems/man/en/chapter_21.htmlhttps://just.systems/man/en/chapter_21.html
+  * We don't need make build deps, etc - just a quick way to wrap tasks.
+ 
+* Shall we count the ways in which the .env file is being used?
```
chriswininger, line 31: I think it probably mostly isn't at this point. It's a way to set environment values that the framework offers and support and if you read documentation on our frameworks examples will often use .env, but in practice I don't think we use it. I prefer setting environment inside the container via the container Environment. For basic local dev you shouldn't need to change anything.

The only time I still imagine you would want to use .env would be if you were a more advanced infinite developers and were doing some unusual flows for debugging, perhaps integration testing against your own auth0 account or a separate s3 instance.

api-server/docker-build.sh
```diff
@@ -1,16 +0,0 @@
-#!/usr/bin/env bash
-
-set -e
```
chriswininger, line 3: I believer your github action scripts rely on this so we would have to change those to move away from this

api-server/justfile
```diff
@@ -0,0 +1,74 @@
+# load environment vars from .env
+set dotenv-load := true
+
+# display this help
+default:
+  just --list
+
+# build the api-server docker image
+build:
+  docker-compose build api
```
chriswininger, line 10: Is there way to have this tagged with a default and pass through optional other tags. That way all local builds have a tag vs release builds

api-server/justfile
```diff
@@ -0,0 +1,74 @@
...
+
+# publish the api-server docker image to docker.io
+publish:
+  docker login
+  docker-compose push api
```
chriswininger, line 15: what tag will this get published as. I don't quite get how the repo is being configured

api-server/Dockerfile
jswank: There isn't much practical difference, but COPY is a much simpler operation and the magic in ADD couldn't be depended on in the past.

api-server/Dockerfile
jswank: no - because there is a .dockerignore directive. the reason to separate these is to make use of the docker build cache more effectively - not refetch dependencies every time a minor edit is made on source code.

api-server/README.md
jswank: Here's the shell script - no extra dependencies needed - https://github.com/ekalinin/github-markdown-toc/blob/master/gh-md-toc

api-server/justfile
jswank: I didn't publish the image, but the docker-compose.yml has these directives:
```
  api:
    build: ./
    image: ${IMAGE_NAME}:${IMAGE_TAG:-latest}
```
I.e. the default tag would be "latest".   The simplist way to override this would be something like

```console
$ IMAGE_TAG=development just build
```

PR #358, opened 2023-01-29
Rework venue address storage

### LINKED ISSUE #318: Rework venue address storage
**Database**
* ✅ Split `address` column into `street`, `city`, `state`, `zip`, `neighborhood`
  * Look into standard ways of handling addresses in Postgres (some ideas [here](https://groups.google.com/g/postgis-users/c/jz56I63JKQU))
* ~Figure out how to reliably wrangle the existing data into the new format~
  * See #350 

**Back end**
* :white_check_mark: Handle new address format

**Front end**
* :white_check_mark: Handle new address format on venue submission
* :white_check_mark: Handle new address format on admin venue interface
* :white_check_mark: Add new [combobox](https://v15.vuetifyjs.com/en/components/combobox/) input field: `State`
  * Suggest Kentucky and the bordering states
* :white_check_mark: Make `City` input into a combobox, too
  * Suggest the cities that are most common in our existing data

rlmead:
Addresses #318 

I'm happy for any feedback, especially on:
1. The selection and ordering of [suggested cities and states](https://github.com/infinite-industries/infinite/blob/318-rework-venue-address-storage/web-portal/components/AddNewVenue.vue#L112)
2. [This little bit](https://github.com/infinite-industries/infinite/blob/318-rework-venue-address-storage/web-portal/components/admin-venue-edit/VenueCard.vue#L266) that updates the soon-to-be-obsolete `address` field when any of its component parts is updated - seems to me like that could be prettier
3. Design/UX

### DIFF
--- a/web-portal/components/AddNewVenue.vue
+++ b/web-portal/components/AddNewVenue.vue
@@ -13,10 +13,10 @@
             </v-flex>
           </v-layout>
 
-          <!-- Venue Address -->
+          <!-- Street Address -->
           <v-layout row wrap>
             <v-flex xs12>
-              <v-text-field label="Address*" v-model="new_venue.address" :rules="[v => !!v || 'Address is required']"></v-text-field>
+              <v-text-field label="Street Address*" v-model="new_venue.street" :rules="[v => !!v || 'Street Address is required']"></v-text-field>
             </v-flex>
           </v-layout>
 
@@ -24,6 +24,13 @@
           <v-layout row wrap>
             <v-flex xs12>
               <v-text-field label="City*" v-model="new_venue.city" :rules="[v => !!v || 'City is required']"></v-text-field>
+            </v-flex>
+          </v-layout>
+
+          <!-- State -->
+          <v-layout row wrap>
+            <v-flex xs12>
+              <v-text-field label="State*" v-model="new_venue.state" :rules="[v => !!v || 'State is required']"></v-text-field>
             </v-flex>
           </v-layout>
 
@@ -74,7 +81,7 @@
         showVenueLoadingSpinner: false, // maybe
         new_venue: {
           name: '',
-          address: '',
+          street: '',
           city: '',
           zip: '',
           neighborhood: '',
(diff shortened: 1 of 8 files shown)


MatthewGidcomb (approved):
Looks good.

web-portal/components/admin-venue-edit/VenueCard.vue
```diff
@@ -224,11 +258,24 @@
...
+            'address': [
+              venue.street,
+              venue.city,
+              venue.state,
+              venue.zip,
+              venue.neighborhood
+            ].filter(a => a).join(', ')
+          }
+        }
+
```
MatthewGidcomb, line 278: I think this is a fine way of handling this, especially since we expect it to be a temporary arrangement. If you're concerned about readability, two things to consider would be

- a comment explaining that `address` is the deprecated composite of the other address fields
- extract the `filter`-`join` construct to another method or utility function with a name like `constructAddressFromFields`

web-portal/components/AddNewVenue.vue
```diff
@@ -85,9 +104,46 @@
...
+        return [
+          'Kentucky',
+          'Illinois',
+          'Indiana',
+          'Missouri',
+          'Ohio',
+          'Tennessee',
+          'Virginia',
+          'West Virginia'
+        ]
```
MatthewGidcomb: Thinking about this a little more and I'm wondering if there needs to be a comment for each of these lists to remind us in a year or two of why we chose the ones we did, in the order we did.

MatthewGidcomb: Or, alternatively, for the benefit of hypothetical future project forkers who won't have the context of today's discussion (although the issue does summarize the reasoning, so maybe I'm overthinking it).

rlmead: i like that idea!

regarding the order of states, i just put them in alphabetical order with Kentucky at the top, but am open to other suggestions. if we took the same approach as we did with cities, we'd narrow the list to KY, OH (Cincinnati & Lorain), and WV (Parkersburg) (i _think_ that list is right based on eyeballing the list of cities, but we can reality-check the list of states with Jason's cleaned-up location data).

chriswininger (approved):

PR #362, opened 2023-02-12
Display full address value (while we still have a separate field for it)

rlmead:
One more little addition to the address work merged with #358


MatthewGidcomb (approved):

chriswininger (approved):
+1, thx

PR #370, opened 2023-02-26
Fix Lat/long button

chriswininger:
The corrects behavior on the venue editor under the admin section.

It now retrieves the correct lat/long


api-server/src/venues/gps.services.ts
```diff
@@ -1,64 +1,41 @@
...
-            // some links are shortened, but will redirect to a full url with the coordinates embedded, use the full url
-            const resolvedRedirectUrl = await this.getRedirectUrl(googleMapLink)
-
-            const [hasCoordinatesInRedirectUrl, redirectCoordinateValues] = this.isUrlWithMatch(resolvedRedirectUrl)
-
-            return hasCoordinatesInRedirectUrl ? this.fromCoordinateValuesString(redirectCoordinateValues) : null
+    async getCoordinatesFromAddress({ street, city, state, zip }: AddressIdentifier): Promise<Nullable<GPSCoordinates>> {
+        if (isNullOrUndefined(MAP_BOX_API_KEY)) {
+            console.warn("attention developer, did you forget to set MAP_BOX_API_KEY")
+            throw new HttpException("No API Token Set", 403)
```
MatthewGidcomb: Is this really a `403` / forbidden condition? It's not that the user isn't authorized, but rather that the server is misconfigured. Seems more like a `500` condition.

chriswininger: yeah I could see that. I was just thinking 403 because auth can't be performed but either way works for me in this case

chriswininger: I feel like this is not the first time I've landed on this post lol https://stackoverflow.com/questions/68559540/http-status-code-for-server-configuration-problem

chriswininger: I suppose auth errors usually indicate the user can do something about it where as 5xx is usually something the dev/admin needs to looks at. I wish there was a more specific 500, but I think you're right 5xx is better. Going with 500 and a more full message

chriswininger: 662f15dc688545276eb5b314fbf06e164948e448

MatthewGidcomb (approved):

rlmead (approved):

Issue #373, opened 2023-03-05
Smaller Submission Form Changes


zoe-triantafilou:
1. Removal of form element/question "is your event..." in-person,online,hybrid. 
2. Change in placeholder text for one form element. Placeholder text in Admission Fee element should read: "Make sure to include multiple admission fees if relevant (adult, child, senior,...)"
3.  Change order of form elements. After the Date/Time and Venue widgets the order of form elements is:
- Event Image
- Admission Fee
- Brief Description
- Full Event Description
- Tags
- Event Website Link
- Ticket Link
- Facebook Event Link
- Eventbrite Link
- Contact Email

Documented at: https://www.figma.com/proto/TYVAMSagx0ufI1DnT9A7ZZ/Infinite-Industries-Submission-Form-2?node-id=52%3A445&scaling=scale-down-width&page-id=3%3A8&starting-point-node-id=52%3A445&show-proto-sidebar=1
Flow: None Selected

zoe-triantafilou:
figure out design for adding online/hybrid option near venue selection widget

PR #374, opened 2023-03-16
Reorganize fields on SubmissionForm

MatthewGidcomb:
I tried filling it out with a real event, and ... I really liked it.

Screenshot:

Screen Shot 2023-03-15 at 21 12 23

references #373


rlmead (approved):
tested locally, looks good 👏🏼

PR #375, opened 2023-03-17
Manage secrets in ansible.

jswank:
This PR does a few tightly integrated things to support managing secret / non-secret information consistently within ansible:

- assign variables in group_vars/<environment>
- use templates + variables to generate the files needed by runtime services
- update the ansible execution scripts to use the environment groups (local, stage, prod) to limit execution
- don't go overboard w/ vars & templates- there are environment specific *files* in nginx/ and docker-files/keys
--- 

I added environment specific secrets to the ansible scripts: you can see these in group_vars.  The .env files are now created via templates, and the scattering of other environment variables in playbooks and invocations are consolidated.

It is possible to do more consolidation since most of the third party services have the same info for staging & prod.  But, because ansible already supports so many places that variables can be defined, I tried to be concise and consistent to make maintenance easier.  Here is the differentiation this PR implements:

- hosts - has variables specific to the execution of ansible
- group_vars/all - lists *all* the variables used by playbooks, usually with placeholder values. Normally these would be defined as defaults on a per-playbook or role basis, but it is important they are listed somewhere and I wanted to minimize changes in this PR.
- group_vars/env/secrets - environment specific secret variables
- group_vars/env/vars - environment specific variables 

This branch has been run against staging and all services restarted following execution there.  There is a backup ~/docker-files in production and the branch is safe to run against it.

Using inventory-based variables lends itself to limiting which hosts playbooks operate against using `-l <environment>` as an invocation mechanism for ansible-playbook rather than using an environment variable as was done previously.  All the associated bash scripts have been updated to reflect this.  Personally, I'd like to ditch the scripts (or move them to a bin directory) to make the repo easier to navigate-  I added a [justfile](https://github.com/infinite-industries/infinite/blob/secrets/ansible/justfile) which effectively replaces them.


ansible/README.md
```diff
@@ -15,36 +14,86 @@ You will need your public key deployed to the host for ssh access
...
 
-Create secret files under secrets-[env]:
+**Setup the Ansible Vault Passphrase**
+There are passwords, secret keys, and other sensitive information required for everything
+to work. These secrets are included in the repo and they are encrypted using
+ansible-vault.  To run ansible-successfully, you will need the passphrase to
+decrypt them.  Ask a team member. Then, run:
+
+```console
+$ echo -n "passphrase" >> .passworf
```
chriswininger: should this be `.password`?

ansible/README.md
```diff
@@ -15,36 +14,86 @@ You will need your public key deployed to the host for ssh access
...
+prod $ docker-compose up -d
+prod $ sudo systemctl restart nginx
+```
+
+### Task: First Time Setup
 
 **These steps only needs to happen once**
 
-After running ansible you will need to ssh into the machine and run:
+1. Add the IP address and other info the appropriate section of the `hosts`
```
chriswininger, line 59: Does this step belong at the beginning of this section, it feels a bit out of place coming after instructions for running ansibe. I think the idea before was that these sections were somewhat linear, with this whole "These steps only needs to happen once" being something that you can usually skip

ansible/README.md
```diff
@@ -15,36 +14,86 @@ You will need your public key deployed to the host for ssh access
...
+
+2. Do the initial install: `ansible-playbook -l staging base_playbook.yml`
+* alternative: `just init staging`
+
+3. Setup certbot (per these [instructions](https://certbot.eff.org/instructions?ws=nginx&os=ubuntufocal).
+
+```
+$ sudo certbot certonly --nginx
+```
+* enter: *staging.infinite.industries,staging-api.infinite.industries* (or *infinite.industries,api.infinite.industries* if this is for prod)>
```
chriswininger: Is the * bookending this here to make it bold, or is it a wild card? I'm a little worried I might copy that if tying to run while not very awake.

chriswininger: Might be worth keeping the link to https://certbot.eff.org/instructions?ws=nginx&os=ubuntufocal somewhere around. There's nothing more fun than having to go back to google and try to guess at where you learned to do something while you're sleepy and prod is down

ansible/README.md
```diff
@@ -15,36 +14,86 @@ You will need your public key deployed to the host for ssh access
...
+will be simplified.
+
 ## Before Running
 
-Create secret files under secrets-[env]:
+**Setup the Ansible Vault Passphrase**
+There are passwords, secret keys, and other sensitive information required for everything
+to work. These secrets are included in the repo and they are encrypted using
+ansible-vault.  To run ansible-successfully, you will need the passphrase to
+decrypt them.  Ask a team member. Then, run:
```
chriswininger, line 26: Can there be only one password?

jswank: All files encypted by ansible-vault use the same password.

ansible/deploy-site_local.sh
```diff
@@ -1,7 +1,4 @@
 #!/usr/bin/env bash
 
-# This is used for deploying to local vms for testing
+ansible-playbook -l local deploy_site_playbook.yml
```
chriswininger, line 3: do we want to loose the ability to change the image version flag?

chriswininger: Probably not a huge loss, but just raising the point

ansible/README.md
chriswininger: guess that's one advantage of git-crypt but it was also a pain in the ass and this is still better than passing around files in slack so :+1:  seems like a good move in the right direction

chriswininger:
Maybe we can add some comments to the reame poinging to docks on some of these different things, ansible-vault, j3, just, ansible.

These all seem like good things that make it easier to tell a clean story with the commands but they also add a bit to the number of things someone new will need to learn before they can effectively makes changes to this layer, something to keep in mind. I'm not overly concerned though. I just think a few links might help call out what's in use and what you should familiarize yourself with

ansible/hosts
```diff
@@ -1,20 +1,8 @@
-[infinite-staging]
-staging.infinite.industries
+[staging]
+staging.infinite.industries ansible_user=infinite
```
chriswininger, line 2: I've gone back and fourth on using ip vs using dns for this sort of thing. DNS sounds more flexible but you can't address if there are dns issues and also you shoot yourself in the foot if you pointed the dns wrong. The IP feels more precise. Thoughts?

ansible/group_vars/staging/vars
```diff
@@ -0,0 +1,6 @@
+---
+env: staging
+infinite_image_version_tag: development
```
chriswininger, line 3: I suppose this might answer my earlier question about deploying different image tags, feels a little less discoverable than passing it in the bash script as somone less familiar with ansible, but probably a better way to do it

ansible/docker-files/web-portal.env.j2
```diff
@@ -0,0 +1,25 @@
+# The environment {dev|staging|prod}
+ENV={{ env }}
+
+# Optional ID to turn on google analytics
+GOOGLE_ANALYTICS_ID={{ google_analytics_id }}
+
+# AWS
+AWS_REGION={{ images_aws_region }}
```
chriswininger, line 8: any reason to have these not correspond 1-1

`AWS_REGION={{ AWS_REGION }}`

ansible/docker-files/api.env.j2
```diff
@@ -0,0 +1,42 @@
...
+
+# The Port to run the application on
+PORT={{ api_port }}
+
+# Slack webhooks are used to send messages about events on the website to a slack channel (they are not required)
+SLACK_WEBHOOK_EVENT_SUBMISSION={{ slack_webhook_event_submission }}
+SLACK_WEBHOOK_VENUE_SUBMISSION={{ slack_webhook_venue_submission }}
+
+# Bitly is used to provided shortened urls for sharing events to social media
+#BITLY_TOKEN="401cbf1121b4859d3a0c8f09b623193eb0233fac"
```
chriswininger: This probably shouldn't be hardcoded here, in fact this is considered a secret I'm pretty sure. We may want to cycle this, though that said I feel like we may have stopped using bitly

ansible/deploy_site_playbook.yml
```diff
@@ -1,32 +1,33 @@
...
+# disable host root access
 
 -
-  hosts: "infinite-{{ lookup('ansible.builtin.env', 'INFINITE_DEPLOY_ENV') }}"
-  vars:
-    os_environment:
-      # used to tell docker-compose what tag to use for the images master/development/etc...
-      - key: INFINITE_IMAGE_VERSION_TAG
-        value: "{{ lookup('ansible.builtin.env', 'INFINITE_IMAGE_VERSION_TAG') }}"
+  hosts: all
```
chriswininger, line 4: how does all work here. I assume it doesn't really deploy to all

chriswininger (approved):
+1

PR #378, opened 2023-03-19
Service management via ansible, systemd.

jswank:
This is the most significant PR of the batch: 
- The docker-compose init is now done by systemd rather than logging into the host and executing docker-compose and / or adhoc scripts.  Upon reboot of the host, the infinite containers will start.
- Anytime configuration for nginx or the docker containers is pushed via ansible, the appropriate services are restarted.  Manually logging in and restarting services is not required.
- An ansible play for pulling the latest (environment-dependent) docker image is included.  No need to log into hosts to do this.
- `just` recipes for getting service status, restarting services, and pulling the latest docker image are included.
- docs are updated to reflect this.


ansible/deploy_site_playbook.yml
```diff
@@ -3,38 +3,74 @@
...
         dest: ~/docker-files/keys/1nfinite.pem
+      notify: Restart docker containers
     - name: Create secrets file for containers
       template:
         src: docker-files/{{ item }}.j2
         dest: ~/docker-files/{{ item }}
       loop:
         - web-portal.env
         - api.env
+      notify: Restart docker containers
```
chriswininger, line 36: Will this cause issues on first deploy with the cert-bot stuff. I forget all the details, I think what happens is if you configure nginx the ways it's supposed to be configured it bombs because cert bot has not created the pem files yet.

At the same time you can't create them because certbot uses nginx to validate when creating the certs so you get stuck in a loop.


Maybe this isn't an issue. This is the second playbook we run so maybe at this time we've already ran certbot, just thinking outloud and double checking :-)

ansible/justfile
```diff
@@ -11,6 +11,27 @@ init env="staging":
 deploy env="staging":
   ansible-playbook -l {{ env }} deploy_site_playbook.yml
 
+# query the status of services for an environment
+status env="staging":
```
chriswininger, line 15: nice, liking this just more and more :-)

jswank: ;)

ansible/deploy_site_playbook.yml
jswank: Restarting the docker containers won't be affected by certbot, but I'll take a closer look at the nginx aspects to understand the impact.  I suspect that addressing it will be more a matter for documentation of initial / manual setup procedures than changing this playbook.

PR #380, opened 2023-03-19
Location info on Cards and in full event views

### LINKED ISSUE #319: Design & rework venue address display
Currently, we don't display any event address information on the cards. We could consider displaying city and/or state somewhere prominent, and perhaps full address information somewhere less prominent.

We might also consider using map link data to display a map, sort of like @jswank does [here](https://observablehq.com/d/3a17b8a83df81b83#cell-346).

Places to consider location data redesign:
* Event card
* Event page
* Venue input
* Admin venue editing

### LINKED ISSUE #359: Fix text alignment on "More Info" and "Add to Calendar" Buttons in Mobile View
Vertical text alignment in mobile view is off by several pixels. 

Quick hack fix: set `class="card-btn add-to-calendar"` attributes to `padding-top:4px `
Need to investigate if there is a more graceful solution.

Screen Shot 2023-02-05 at 7 42 37 AM

Digi-D:
Show city location only from the specific event info on the cards on the main page. 
Show full address from the specific event on the event page

### DIFF
--- a/web-portal/components/EventCard.vue
+++ b/web-portal/components/EventCard.vue
@@ -6,6 +6,9 @@
         <div class="image-container">
           <nuxt-link :to="eventLink">
             <div class="image-surface" :style="backGroundImage"></div>
+            <div class="image-location-city-container">
+              <div class="image-location-city">LOCATION</div>
+            </div>
           </nuxt-link>
         </div>
         <div class="info-container">
@@ -249,6 +252,17 @@
 
     border-top-left-radius: 10px;
     border-top-right-radius: 10px;
+  }
+  .image-container .image-location-city-container {
+    background-color: red;
+    position:relative;
+    top: -20px;
+    text-align: right;
+  }
+
+.image-location-city-container .image-location-city {
+    background-color: green;
+    max-width: 120px;
   }
 
   .image-container:hover .image-surface {
--- a/web-portal/components/EventCard.vue
+++ b/web-portal/components/EventCard.vue
@@ -7,7 +7,7 @@
           <nuxt-link :to="eventLink">
             <div class="image-surface" :style="backGroundImage"></div>
             <div class="image-location-city-container">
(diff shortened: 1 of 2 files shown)


web-portal/components/EventView.vue
```diff
@@ -188,6 +193,26 @@
...
+          // (but don't include state w/o city)
+          if (venue.city) {
+            addr += ' ' + venue.state ? `${venue.city}, ${venue.state}` : venue.city
+            // and only include zip if city, state is known
+            if (venue.zip) addr += ' ' + venue.zip
+          }
+        }
+
+        return addr
+      },
```
MatthewGidcomb: It looks like we forgot to actually wire this in on the template -- it's still building it up manually (lines 17-19).

web-portal/components/EventCard.vue
```diff
@@ -141,6 +144,7 @@
           this.calendar_event.tags.includes('condition:sold-out')
       },
       isRemote: function () {
+        console.log(this.calendar_event.title + ' - ' + JSON.stringify(this.calendar_event.tags))
```
chriswininger: we should delete this before we check in. It's not great to leave in console logs, they can degrade performance if nothing esle

chriswininger (approved):
I mostly yield to Mathew's approval for frontend but +1 from me with one minor suggestion

web-portal/components/EventView.vue
MatthewGidcomb: That was my mistake -- fixed

web-portal/components/EventCard.vue
MatthewGidcomb: Removed, thanks for flagging.

MatthewGidcomb (approved):

MatthewGidcomb:
Will close #319 and #359.

PR #383, opened 2023-03-20
Service.mgmt 

jswank:
This is the most significant PR of the batch:

The docker-compose init is now done by systemd rather than logging into the host and executing docker-compose and / or adhoc scripts. Upon reboot of the host, the infinite containers will start.

Anytime configuration for nginx or the docker containers is pushed via ansible, the appropriate services are restarted. Manually logging in and restarting services is not required.

An ansible play for pulling the latest (environment-dependent) docker image is included. No need to log into hosts to do this - there are just recipes for getting service status, restarting services, and pulling the latest docker image are included.
docs are updated to reflect this.


ansible/README.md
```diff
@@ -37,19 +53,45 @@ $ just cache-pass
 
 ## Running
 
-### Common Task: Site Deployment
+### Using `just`
```
chriswininger, line 56: nice, thx for adding this

chriswininger (approved):
This is a great improvement, awesome!

PR #384, opened 2023-03-20
Image version definition

jswank:
The docker-compose configuration relies on an environment variable, INFINITE_IMAGE_VERSION_TAG, to indicate which version of an image should be used for docker-compose.  This variable will now be defined in a .env file used by docker-compose rather than /etc/environment. 

Per discussion w/ Chris: overriding this tag can be a common activity.  To support this more readily, the PR also adds the ability to pass arbitrary arguments to ansible-playbook when using just.  For instance:
```console
$ just deploy flags='-e image_version=master'  # use the master container in staging
# just deploy prod flags='-e image_version=development'  # use the dev container in production
```


chriswininger (commented):
Looks, good the only thing I don't quite understand is how the values set inside .env get set in the host environment for docker-compose to pickup.

I see how .env.j2 get turned into a .env file to be copied but where does it get used

chriswininger (approved):
+1

PR #385, opened 2023-03-25
S3 updates

jswank:
update S3 credentials for image upload.  correct region. remove superflous s3 url definition.

This is currently running in staging.


chriswininger (approved):
+1

PR #399, opened 2023-05-07
Add the MAP_BOX_API_KEY to api-server config

jswank:
This sets MAP_BOX_API_KEY for the api-server.

I've run against staging and does the expected - adding the Mapbox API token to docker-files/
api.env.

```console
$ just deploy staging
```


chriswininger (approved):
Awesome thanks for cranking this out

PR #424, opened 2023-07-01
Add additional dev tooling and codespaces automation

jswank:


.devcontainer/Dockerfile
```diff
@@ -1,5 +1,12 @@
 FROM mcr.microsoft.com/devcontainers/javascript-node:0-16
 
 RUN apt-get update \
-  && apt-get install -y postgresql-client tmux zsh neovim
+  && apt-get install -y postgresql-client tmux zsh neovim ansible
 
+RUN curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin
```
chriswininger, line 6: I'm fine with this especially for dev containers but don't love downloading straight from urls in docker containers. Sometimes it's needed, but if there's any other options, apt, checkend the binary etc... probably preferable. This has definitely bit me, but I definitely continue to do it all the time :-)

jswank: I don't love it either and am open to reasonable alternatives.  Both casey/just and aws cli are installed this way... and it is the recommended approach for the image used in the devcontainer (mcr.microsoft.com/devcontainers/javascript-node:0-16 i.e. Debian 11 / "bullseye").   

Options I've considered include:
1. The curl-pipe-install script in this PR. Pros: it's the recommended approach, and simple to understand + update / fix down the road.  Cons: questionable aesthetics / security / resilience to upstream changes.  I chose this route with the rationale that if / when it breaks, then Codespaces users can update the image appropriately. 
2. Making use of the [prebuilt MPR](https://docs.makedeb.org/prebuilt-mpr/introduction/) for casey/just and the awscli package.  Pros: uses apt-get to install packages. Cons: requires setup of additional apt repo (for prebuilt MPR), uses an older / non-current AWS cli, and is probably just as fragile as any other mechanism.
3. Using a custom development image.  Pros: can use Alpine or another OS that makes installing modern tooling trivial.  Cons: any / all the associated work of configuring the right version of Node / etc for development purposes.

PR #426, opened 2023-07-02
Draft: Add pagination to verified events endpoint

chriswininger:
Adds pagination support to events/verified

This endpoint returns all (not just current) verified events. Historically this endpoint has not worked because it returns too much data. We don't actually use it currently but it could be useful in the future and will likely be needed for many third party integrations.

By adding pagination this endpoint should now be usable.

* The default page is 1 and the default pageSize is 20, but these can be changed by the user.
* The pagination can also be paired with a pre-filter on a single or multiple tags
* Only verified events are returned

I have dropped support for the includes tag which let you specify if nested models like venue or date_times are included. To do the filtering and sorting we need to fetch them in all cases and given that are models are not super large they do not seem necessary.

Since no we are not actually leveraging this endpoint and it hasn't been working the risk of breaking changes is low. I would like to possibly generalize some of these changes to our other endpoints next.


api-server/src/events/dto/eventDTO.ts
```diff
@@ -65,6 +65,9 @@ export default class EventDTO {
   @ApiProperty({ example: EXAMPLE_DATE })
   createdAt: Date;
 
+  @ApiProperty({ example: false })
+  multi_day: boolean;
+
```
chriswininger, line 70: ^ This was a bug I found while testing. We have not been serializing multi_day when we fetch

MatthewGidcomb: Is this property even used? I think it may be an old column from an earlier format.

It's definitely not the column affected by the `category` radio group (single/multi-day event, gallery show, etc).

api-server/src/events/events.controller.ts
```diff
@@ -37,6 +37,12 @@ import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
 import { eventModelToEventDTO } from './dto/eventModelToEventDTO';
 import EventDTO from './dto/eventDTO';
 import { ENV } from '../constants';
+import { ApiImplicitQuery } from '@nestjs/swagger/dist/decorators/api-implicit-query.decorator';
```
MatthewGidcomb, line 40: Is this how this is intended to be imported? Seems pretty messy by Nest's normal standards.

api-server/src/events/events.controller.ts
```diff
@@ -103,21 +109,55 @@ export class EventsController {
...
+  @ApiImplicitQuery({
+    name: 'page',
+    description: `the requested page (default ${EVENT_PAGINATION_DEFAULT_PAGE})`,
+    example: 1,
+    required: false,
+    type: Number,
+  })
+  @ApiImplicitQuery({
+    name: 'pageSize',
+    description: `the number of events to user per page (default ${EVENT_PAGINATION_DEFAULT_PAGE_SIZE})`,
```
MatthewGidcomb:
```suggestion
    description: `the number of events included per page (default ${EVENT_PAGINATION_DEFAULT_PAGE_SIZE})`,
```

api-server/src/events/events.service.ts
```diff
@@ -56,6 +67,71 @@ export class EventsService {
...
+        },
+      );
+
+      // Fill back in nested models date_times and venues
+      const dateTimes = await this.dateTimeVenueModel.findAll({
+        where: {
+          event_id: {
+            [Op.or]: paginatedRows.map(({ id }) => id),
+          },
+        },
```
MatthewGidcomb, line 114: I don't know how Sequelize works under the hood, so maybe this isn't even possible, but would this make more sense as an `IN` condition, e.g. `event_id IN (id1, id2, ...)`?

MatthewGidcomb (approved):

api-server/src/events/events.controller.ts
chriswininger: this seems to be the only place to import it. I agree it's a bit of an odd path

api-server/src/events/dto/eventDTO.ts
chriswininger: probably right,  this makes it work consistently so I say we leave this fix, but in another pr delete it properly from the db through the code

PR #437, opened 2023-08-20
Move Authenticated Endpoints

chriswininger:
Moved the endpoints for getting all events and for getting non-verified endpoints to the authenticated controller. This is a bit of house keeping. The intent was that all endpoints that require authentication would be under a /authenticated route. This is for external communication and it also makes it so we can turn auth at the controller level rather than on each endpoint.

I also backfilled some tests


api-server/test/fakers/event.faker.ts
```diff
@@ -132,4 +132,65 @@ export async function createEvent(
...
+  venueModel: typeof VenueModel,
+  datetimeVenueModel: typeof DatetimeVenueModel,
+  numEvents = 30,
+  overrides: EventModelConstructorProps = {},
+  baseTime = new Date(),
+): Promise<[EventModel[], Date]> {
+  const events: EventModel[] = [];
+
+  for (let i = 0; i < numEvents; i++) {
+    const [newEvent, newBaseTime] = await createEventInPast(
```
MatthewGidcomb, line 146: It's a little confusing that `createListof**Future**EventsInChronologicalOrder` delegates to another helper called `createEvent**InPast**`.

MatthewGidcomb (approved):

api-server/test/fakers/event.faker.ts
chriswininger: ah, indeed lol, probably need to rethink those names

rlmead (approved):

Issue #409, opened 2023-06-06
Clearing description on category "other" event produces duplicate category tag


MatthewGidcomb:
Reproduction steps:

- Select the "Other" category and enter some text in the "What type of event are you putting on?" input field.
- Clear the input field.
- Enter new text into the field.
- Use the Vue dev tools to inspect the Submission form's state or submit the event and inspect the resulting event.
- The `tags` column should contain `category:other:<the text you entered in the field>` but instead it contains that plus an additional `category:other` tag.

MatthewGidcomb:
_Should_ be fixed by #450, which completely reworked how category is stored such that it shouldn't be possible to duplicate the info. We'll check before closing for good.

MatthewGidcomb:
Still need to follow up on confirming this is fixed.

MatthewGidcomb:
Confirmed

PR #450, opened 2023-10-01
Rework tags (fixed tests, conflicts resolved)

### LINKED ISSUE #390: Track category, condition, and mode in their own columns
Move `category`, `condition`, and `mode` information out of `tags` in events and making it its own column.

Datetimes for gallery shows (with start_date and end_date vastly far apart) are a somewhat different use case from our usual conception of `datetime_venue`. Theater shows and musical events are also special.

Consider adding a "type" field to the `datetime_venue` table. Special events for galleries could be labeled with `opening` and `closing`.

rlmead:


api-server/migrations/20230416142015-add-columns-for-tags.js
```diff
@@ -0,0 +1,128 @@
...
+            type: Sequelize.ARRAY(Sequelize.STRING),
+            allowNull: true
+          }, { transaction: t }),
+        queryInterface.addColumn(
+          'events',
+          'mode',
+          {
+            type: Sequelize.STRING,
+            allowNull: true
+          }, { transaction: t })]).then(queryInterface.sequelize.query(`
```
MatthewGidcomb: Should `category` and `mode` be non-nullable?  When we first added the frontend support we wrote a migration that back-filled existing events, so we _should_ have data for all events.

web-portal/components/SubmissionForm.vue
```diff
@@ -410,13 +390,13 @@
...
       const new_event = this.$store.getters.GetCurrentEvent
       this.calendar_event = Object.assign({}, new_event, {
+        condition: new_event.condition ? new_event.condition.map(t => t) : [],
         date_times: new_event.date_times.map(dt => ({ ...dt })),
         tags: new_event.tags ? new_event.tags.map(t => t) : []
       })
     },
     methods: {
       UpdateEvent: function () {
-        console.log(this.calendar_event)
```
MatthewGidcomb, line 419: Thank you for cleaning this up -- we should have done it a long time ago.

web-portal/components/SubmissionForm.vue
```diff
@@ -588,49 +565,33 @@
...
+          return (this.calendar_event && this.calendar_event.category) ? this.calendar_event.category.split(':')[0] : ''
+        },
+        set (newValue) {
+          this.calendar_event.category = newValue
+        }
+      },
       eventCategoryOther: {
         get () {
-          const tag = this.calendar_event && this.calendar_event.tags.find(tag => /^category:other/.test(tag))
+          const tag = (this.calendar_event && /^other/.test(this.calendar_event.category)) ? this.calendar_event.category : false
```
MatthewGidcomb, line 582: This definitely works as-is, but I wonder if the regex would be safer / more future-proof if it looked for `/^other:/`.

rlmead: aaahhh interesting! i like it, except that there's one valid case that it doesn't work for: plain old `other` with no further specification. while we do have the prompt for people to tell us what they mean by `other`, they're not **required** to add more info there. so as things currently stand, that's a valid category label, and when i test this out locally it works as i expect it to. thoughts?

(in case you wondered what's currently in the db, like i did) we currently have four events in the db that use the `other` category option. only one of them has the underspecified `category:other` tag, and it's a duplicate (which reminds me that we also need to do [this little bit of database cleanup](https://1nfinite1ndustries.slack.com/archives/C1DUR2VNW/p1685894279849259)).
```
infinite-api=# select id, tag from (select id, unnest(tags) as tag from events) asdf where tag like 'category:other%';
                  id                  |                                     tag
--------------------------------------+------------------------------------------------------------------------------
 7e875f02-22e0-4a2e-8823-fb345b87f878 | category:other:Ticket lottery
 7e875f02-22e0-4a2e-8823-fb345b87f878 | category:other
 6d40e65c-6016-4b42-beb7-7371662e00ff | category:other:Watch-on-demand
 f7102989-3c67-4943-9123-b71acf26037c | category:other:Recurring open mic the first and third Tuesday of every month
 c1d2c113-1d43-4487-a007-c17045e7916b | category:other:Open auditions
(5 rows)
```

api-server/migrations/20230416142015-add-columns-for-tags.js
rlmead: gooood question - i like that idea. this might require further investigation though.

poking in the latest database dump (as it is before running this migration) i see 3927 events total:
- 2231 of these have no category tag
- 1931 of these have no mode tag

[this](https://infinite.industries/events/d8bcbd69-bdf0-4c25-ae71-7fed2cea3f30) is the most recently-created of these events, and it was created on August 25 2022. we have events _without_ category and/or mode going all the way back to 2018. the earliest event _with_ a mode tag was created in August 2019, and the earliest event _with_ a category tag was created in April 2020.

so either we lost some old data in August 2022, or we made a change around that time that ensured that `category` and `mode` would always be stored. 🤔🤔

MatthewGidcomb: Ah, that's interesting. I thought we had run a migration or script to populate old events with category and mode tags, but maybe I'm misremembering. Or did we accidentally blow that away with the accidental production DB deletion?

web-portal/components/SubmissionForm.vue
MatthewGidcomb: Ah! That's probably why it didn't have the `:` initially, and I just forgot about it. When I wrote this comment I was thinking it would be `category:other:` if no description was provided.

MatthewGidcomb: Might also be related to #409.

api-server/migrations/20230416142015-add-columns-for-tags.js
MatthewGidcomb: Based on discussion in today's meeting, and looking back at the initial implementation of mode and category (#289), it looks like we bailed on back-filling these values except for events that had the `remote` and `online-resource` tags from the COVID-era implementation.

MatthewGidcomb (approved):

Issue #436, opened 2023-08-17
All the images are uploaded as a JPEG regardless of the actual file type


Digi-D:
Currently, all of the images are uploaded with .jpg suffix regardless of their actual file type. 

Doesn't break since the S3 bucket returns it as "application/octet-stream" MIME type but needs to be changed to the appropriate suffix or avoid suffixes all together.

MatthewGidcomb:
Cause of hard-coded `.jpg` suffix:

https://github.com/infinite-industries/infinite/blob/5f3033c1292ff6483d432796a074610c0c7a2287/web-portal/internal-api/upload-images.js#L13-L18

MatthewGidcomb:
Fixed by #454

Issue #356, opened 2023-01-29
Downsample Images


chriswininger:
We are currently storing event images exactly as submitted. Some our quite quite large. This makes site loading slower and costs us money. We should down sample these to some target range.

**Acceptance Criteria:**
* Consider how we down sample, do we ensure a certain file size or a certain image size
* Do we need to restrict the input a bit more to make this possible
* Do we have a super down sampled image for the thumbnails vs the event details page

Note: There is a broader question about letting users resize and crop images to fit our format, but for the purposes of this ticket that is not required. Think more about an easy win to prevent 13 MB images getting onto the home page

MatthewGidcomb:
[Here](https://github.com/infinite-industries/infinite/blob/development/web-portal/internal-api/upload-images.js#L46-L97)'s where image uploads are handled. This request handler delegates to a separate uploader class (implementations [here](https://github.com/infinite-industries/infinite/tree/development/web-portal/internal-api/upload-managers)); the specific implementation is chosen at server start based on what's set in the environment.

PR #454, opened 2023-10-30
#356 Part I -- Move Image Upload API

chriswininger:
I am preparing to implement a first pass of basic image per-processing which will make sure we do not store images that are too large. As an initial first step I have moved image processing out the web-portal and into the API-Server where we are supported by the more robust backend functionality that NextJs provieds.

This inadvertanly addresses issue #436. In the re-write I have not ported the bug :-)


api-server/src/constants.ts
```diff
@@ -7,8 +7,10 @@
  */
 import fs from 'fs';
 import path from 'path';
-import { NullableOrUndefinable } from './utils/NullableOrUndefinable';
+import { Nullable, NullableOrUndefinable } from './utils/NullableOrUndefinable';
 import { isNullOrUndefined } from './utils';
+import * as process from 'process';
```
MatthewGidcomb: Why does this need to be imported?

api-server/src/uploads/uploads.controller.ts
```diff
@@ -0,0 +1,59 @@
...
+  ): Promise<UploadsResponse> {
+    const imagePath = await this.uploadsService.saveImage(file);
+    return new UploadsResponse({ imagePath });
+  }
+}
+
+function getImageUploadValidators(): ParseFilePipe {
+  return new ParseFilePipeBuilder()
+    .addFileTypeValidator({
+      fileType: /image\/(jpe?g|png|webp|tif|tiff|bmp)$/i,
```
MatthewGidcomb: Does this support `image/*`? That's what the input on the form uses for `accept` (per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers) it should accept any image format).

(I wish Nestjs had API docs to supplement the higher-level ones.)

chriswininger: yeah this was quite frustrating to get right what I'm matching on are mime types such as image/jpeg, image/png. It's just a regular expression so you could match against image/*, that would be `/image\/.*$/i` I believe.

If that's what we're allowing in the form that would make sense though long term we should probably lock this down more. I doubt we want to support any image format

api-server/src/constants.ts
chriswininger: do you mean the line below this? process is being used, though the import below this `import isNotNullOrUndefined from './utils/is-not-null-or-undefined';` is unused and I will remove it

chriswininger: 86b80a3dbd05d542c4145ac92684e08f184a8e35

chriswininger: 🤔 maybe though you did mean process looks like it an work without importing it, guess it's special, doesn't seem to matter either way

chriswininger: removed 253ceecb7727632ef08d1ab92774ce4343a7bba9

api-server/src/uploads/uploads.controller.ts
chriswininger: loosened up the requirements 9bf391a22a4a612f87b00a114fc108ff9f614b97

PR #459, opened 2023-11-20
Issue-356: Downsample Images -- Part II

chriswininger:
This will check if the incoming image is greater than 1920 pixels wide, if so it resizes down to 1920. Regardless of width it will save the image from the incoming format to a compressed webp format. With some quick local testing I'm seeing images from my phone drop from 3-9megs down to a few hundred kb. They still look totally fine to my untrained eye even on my fancy super high ultra whatever we call it these monitor. This is especially true because even in the full event view the image is not allowed to fill the screen.

A better version of this would actually save at least 2 images in mind, for example we could save an even smaller version for the tiles. I am considering that change for a followup, but that will require migrations and more code changes. This version should go along way towards improving our standing with google and mobile load times using only small changes


api-server/src/uploads/uploads.service.ts
```diff
@@ -11,6 +11,10 @@ import { PATH_TO_LOCAL_EVENT_IMAGE_UPLOADS } from './uploads.module';
 import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
 import { isNullOrUndefined } from '../utils';
 import { S3 } from '@aws-sdk/client-s3';
+import sharp, { Sharp } from 'sharp';
+
+const IMAGE_EXTENSIONS = 'webp';
```
MatthewGidcomb: Nitpick but since this is being used later to set the extension, shouldn't it be `IMAGE_EXTENSION`?

api-server/src/uploads/uploads.service.ts
```diff
@@ -25,28 +29,45 @@ export class UploadsService {
...
     const subPath = 'event-images';
 
     return new Promise((resolve, reject) => {
-      const imageName = this.generateNewImageName(img);
+      const imageName = this.generateNewImageName();
 
       // wx flag mitigates the possibility of clobbering an existing file
       // (will error out on write)
       fs.writeFile(
         join(PATH_TO_LOCAL_EVENT_IMAGE_UPLOADS, imageName),
```
MatthewGidcomb, line 69: Is this path in `.gitignore`?

chriswininger: PATH_TO_LOCAL_EVENT_IMAGE_UPLOADS?

I believe so, though it's kind of weird because depending on how things are ran it  often actually pops up under dist (also in .gitignore, I had a fair bit of trouble with this actually).

That specific comment is probably copy/paste from the old code in the nuxt probject, but still looks true. This is all for local test, so not a big deal either way I think, but I'll stand by the behavior

api-server/src/uploads/uploads.service.ts
chriswininger: yep, fixed bd3116e334f3cc2b26975ecb6eb32a8b6ad4cc62

Issue #469, opened 2024-01-21
intermittent test failure


chriswininger:
I've seen the test '/verified exclude unverified events' fail in ci/cd a few times. It fails because one of the events does not have a start date later than it's predecessor.

This seems like it could indicate a legitimate edge case problme

chriswininger:
Started experimenting on local:

* Ran the test in isolation 10 times in a row and did not see a failure
* Ran the whole suite 10 times and saw it fail in the exact way as CI/CD once
* Created a script as follows to simplify re-runs

```
npx jest --config ./test/jest-e2e.json --runInBand -t "Events API./verified exclude unverified events"
```

ran 20 times

```
for it in {1..20}; do ./caw-run-single-test.sh; done
```

-> No failures

* Ran whole suite 20 times in same way:

```
npx jest --config ./test/jest-e2e.json --runInBand ./test/events.e2e-spec.ts
```

-> 3/20 failures

[failure-example.txt](https://github.com/infinite-industries/infinite/files/14002399/failure-example.txt)

Adding logging to dump all the events when it fails:

PR #472, opened 2024-02-01
Issue #469 - Fix Flaky Test

chriswininger:
We need to sort on the first_start_time a second time after joining our page
back to the events table becuase the order that comes out of the merge is not
guaranteed to be sorted the same as in it was inside the common expression table

basically we were sorting the overall list correctly and guaranteeing each page would always have
the same things on it but that the things within each page would always be in the same order.

Applying this sort correctly revealed a second test which was wrongly assuming online resources, which
have no date_times would come at the end of the page when in fact they will always come at the beginnning
of the page.

To further tighten this up I added a secondary sort on created_at so that not only will online resources always
appear at the beginning of the page, but also in the same order

I left in place a new test based on data which produced durring randomized runs that originally highlighted this
problem

### DIFF
--- a/api-server/src/events/dto/create-event-request.ts
+++ b/api-server/src/events/dto/create-event-request.ts
@@ -87,7 +87,7 @@ export class CreateEventRequest {
 
   @ApiProperty({ example: 'radio-mc-radio-station' })
   @IsOptional()
-  reviewed_by_org: boolean;
+  reviewed_by_org: string;
 
   @ApiProperty({
     example: [{ start_time: EXAMPLE_START_DATE, end_time: EXAMPLE_END_DATE }],
--- a/api-server/src/events/events.service.ts
+++ b/api-server/src/events/events.service.ts
@@ -89,6 +89,11 @@ export class EventsService {
       );
 
       // Sort events by the first start_time and apply pagination
+      // Note, we have to sort by first_start_time in our common table expression to apply pagination correctly, but we
+      // also have to sort again over our page in the final join with events, because the order after joining is not
+      // guaranteed to stay the same.
+      // We've also added a secondary order by on created_at to ensure that events with with not start_time like
+      // online resources at least sort consistently
       const paginatedRows: EventModel[] = await this.sequelize.query(
         `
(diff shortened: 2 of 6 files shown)


api-server/src/events/dto/create-event-request.ts
```diff
@@ -87,7 +87,7 @@ export class CreateEventRequest {
 
   @ApiProperty({ example: 'radio-mc-radio-station' })
   @IsOptional()
-  reviewed_by_org: boolean;
+  reviewed_by_org: string;
```
chriswininger, line 90: this was something else I noticed while working on this, this type was simply wrong

api-server/src/events/events.service.ts
```diff
@@ -89,18 +89,24 @@ export class EventsService {
...
                                                  LEFT OUTER JOIN datetime_venue dv on events.id = dv.event_id
                   ${whereClause}
               GROUP BY (events.id)
-              ORDER BY first_start_time DESC
+              ORDER BY first_start_time DESC, "createdAt" DESC
               OFFSET ${(requestedPage - 1) * pageSize} LIMIT ${pageSize} )
               SELECT events.*
               FROM compressed_event
                        JOIN events ON events.id = compressed_event.id
+                       ORDER BY compressed_event.first_start_time DESC, "createdAt" DESC
```
chriswininger, line 109: this addtion of  `ORDER BY compressed_event.first_start_time DESC` here on the outer fetch is the main fix here, adding in `"createdAt" DESC` is some icing on the cake to further lock in the order

MatthewGidcomb: Why are these sorts `DESC`? I'm sure this is right, given we haven't noticed any issues, but I would have assumed we'd want to sort by ascending dates?

MatthewGidcomb: I get it now -- this isn't the current-events endpoint, it's _all the events_. Makes more sense, for the moment, to get more recent stuff first vs events from 6+ years ago.

MatthewGidcomb (approved):

Issue #473, opened 2024-02-02
api to detect when a new event possibly exists already


chriswininger:

MatthewGidcomb:
Implemented by #475 and #480; still more to do but separate tickets

PR #475, opened 2024-02-10
Issue #473 -- Introduce Duplicate Detection Endpoint

chriswininger:
This introduces a strategy to detect when you are about to submit a duplicate event.

It's driven off of start times and venue ids.

We check to see if there are other events that have the same start time at the same venue.

If more than 50% of the proposed times start at the same time and location as an existing event then we treat that event as a possible duplicate of your event and return data back to the client about it.

Also included is a small one line change to our logger to help us get full traces for easier debugging in production.

image


api-server/src/logging/ExceptionLogger.ts
```diff
@@ -13,6 +13,8 @@ export class ExceptionLogger extends BaseExceptionFilter {
       : 'unknown method';
 
     console.info(`Handled Exception (${url}, ${method}): "${message}"`);
+    console.error(exception);
+
```
chriswininger, line 17: sneak sneaked this in here. This will make sure when there's an exception we get a full stack trace along with the basic message. I think this will be useful for debugging production issues.

i at least want to experiment with having this on. If it gets too noise we can take it back

api-server/README.md
```diff
@@ -71,6 +71,17 @@ $ npm run test:cov
 local: http://localhost:3003/api/
 prod:  https://api.infinite.industries/api/
 
+## API Design Philosophy and Best Practices
+
+* While database models use snake for fields the API should generally use camel case.
+* Exceptions have been made for this where the data directly represents a database entity.
+  * This is not deal, we should not directly expose database model definitions via the api, but shortcuts
+    have been take.
```
MatthewGidcomb:
```suggestion
  * This is not ideal, we should not directly expose database model definitions via the api, but shortcuts
    have been taken.
```

api-server/src/events/existing-event-detection-controller.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@nestjs/common';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import ExistingEventDetectionService from './existing-event-detection-service';
+import { VERSION_1_URI } from '../utils/versionts';
+import { ApiOperation, ApiResponse, ApiTags } from '@nestjs/swagger';
+import ExistingEventDetectionParameters from './dto/existing-event-detection-parameters';
+import ExistingEventDetectionResults from './dto/existing-event-detection-results';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require('dotenv').config();
```
MatthewGidcomb, line 17: Why does this need to be called here (vs at the application entry point)?

chriswininger: I need to investigate that. To be honest I don't remember and I'm not sure if it does, but it's in all the other controllers. I seem to recall there was a weird issue with making that gets initialized. It wasn't really integrated into the nest flow that well.

I think it's fairly likely this is not needed, but I was afraid to leave it out.

I created this issue to look into it https://github.com/infinite-industries/infinite/issues/476

PR #480, opened 2024-02-17
Issue #473 -- Existing Event Warning On Submission Form

chriswininger:
If you enter an event with the same start time at the same location you will get an existing event warning on the submission form. For events with multiple start times this is driven by percentage of matching start times.

For now this warning will only be shown to users that are logged in as admins, effectively restricting it to our team.

Screenshot from 2024-02-17 12-20-16

[Demo Video](https://youtu.be/llULEYzqIBI)

### DIFF
--- a/web-portal/components/ExistingEventDetectionAlert.vue
+++ b/web-portal/components/ExistingEventDetectionAlert.vue
@@ -1,12 +1,12 @@
 <template>
-  <div class="existing-event-alert">
-    <h3>Existing Event Detected</h3> <span class="existing-event-alert__icon">⚠️</span>
+  <div class="existing-event-detection-alert" v-show="isShown">
+    <h3>Existing Event Detected</h3> <span class="existing-event-detection-alert__icon">⚠️</span>
 
     <p>Someone has entered a similar event starting around the same time at the same location.</p>
 
     <div>Possible Matches:</div>
     <ul>
-      <li v-for="candidate in duplicateEventsByStartTime.candidateEvents" :key="candidate.url">
+      <li v-for="candidate in getCandidates()" :key="candidate.url">
         <a :href="candidate.url">
           {{ candidate.title }}
         </a>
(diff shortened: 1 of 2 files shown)


web-portal/components/SubmissionForm.vue
```diff
@@ -517,14 +523,52 @@
...
+          })
+          .catch((err) => {
+            console.error('Error performing duplicate detection on start times: ', err)
+          })
+      },
+      shouldShowExistingEventDetectionByStartTime: function () {
+        return this.duplicateEventsByStartTime !== null &&
+          this.duplicateEventsByStartTime !== undefined &&
+          this.duplicateEventsByStartTime.isLikelyExisting
+      },
```
MatthewGidcomb: This could be a computed property instead of a method.

web-portal/components/ExistingEventDetectionAlert.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <div class="existing-event-detection-alert" v-show="isShown">
```
MatthewGidcomb: `v-if` instead of `v-show` would suppress rendering altogether (which is generally my preference in a case like this but it's just a preference).

chriswininger: sweet that's what I wanted, but didn't understand how to specify in vue, will change

chriswininger: 1aad33ca29cb31e60a0b40ab558690885e142d78

web-portal/components/SubmissionForm.vue
chriswininger: 1aad33ca29cb31e60a0b40ab558690885e142d78

chriswininger: 1aad33ca29cb31e60a0b40ab558690885e142d78

MatthewGidcomb (approved):

PR #488, opened 2024-02-25
Removing Vutify From DateTimePicker

chriswininger:
I am making use of a library called Flatpickr to replace the date chooser. It is pretty much a drop in replacement.

I am using their 2.x version which supports vue2. They have a 3.x version for when we switch to vue3. I believe the migration should be painless.

I am also replacing the v-button elements

Look ma no vutify

image


web-portal/components/DatePicker.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div>
+    <flat-pickr readonly="true" v-model="date" :config="config"></flat-pickr>
+  </div>
```
MatthewGidcomb: Is the wrapping `div` necessary?

e2e-tests/integration/eventEditingTest.js
```diff
@@ -130,9 +130,9 @@ context('Event editing:', () => {
 
     // add a new datetime
     // advance calendar one month and select the 10th
-    cy.get('#cal-container .v-date-picker-header > button:last-child').click()
+    cy.get('#cal-container .flatpickr-next-month').click()
     cy.wait(1000) // wait for calendar animation to complete
```
MatthewGidcomb, line 134: We can deal with this later, but the `wait` might not be necessary with Flatpickr -- it was just there because we couldn't figure out a way to wait on the Vuetify one's animation.

web-portal/components/DatePicker.vue
```diff
@@ -0,0 +1,45 @@
...
+    model: {
+      event: 'change'
+    },
+    data: function() {
+      return {
+        date: null,
+        config: {
+          inline: true,
+          enableTime: false,
+          minDate: 'today',
```
MatthewGidcomb: Small thing we should investigate: with the current setup we allow any selection (including in the past) on the admin page.

web-portal/components/DateTimePicker.vue
```diff
@@ -9,13 +9,7 @@
...
     <div class="time-date-control-wrapper">
       <div>
-        <v-date-picker
-          v-model="picker"
-          color="gray lighten-1"
-          :allowed-dates="AllowedDates"
-          :no-title="true"
-          :dark="true"
-        />
+        <date-picker @change="dateChanged" />
```
MatthewGidcomb: How does this get populated with the selection when editing a datetime?

web-portal/package.json
```diff
@@ -39,7 +39,9 @@
     "nuxt-clipboard2": "^0.2.1",
     "slack-notify": "^2.0.2",
     "uuid": "^9.0.0",
+    "v-calendar": "^2.4.2",
     "vue-clipboard2": "^0.3.0",
+    "vue-flatpickr-component": "^8.1.8",
```
MatthewGidcomb: Is `v-calendar` required w/ flatpickr?

web-portal/components/DatePicker.vue
chriswininger: I was wondering that myself. I hadn't gotten around to testing what happens without it

e2e-tests/integration/eventEditingTest.js
chriswininger: yeah wondered about that, but didn't want to change more than I needed

web-portal/components/DatePicker.vue
chriswininger: oh I see, specifically for the admin page? That would be easy enough to implement if we want that

web-portal/components/DateTimePicker.vue
chriswininger: not sure I follow, I guess the answer is it doesn't. It goes the other direction. You pick a date in this component which fires the dateChanged and that sets a new time. Is there any reason this needs to work the other way?

web-portal/package.json
chriswininger: hope not, I think I left that in there from my earlier experiments

chriswininger: 516fdb169070de85969595c2c65016f684e56999

e2e-tests/integration/eventEditingTest.js
chriswininger: removed 26c88da4454a1de00277cade533e92664483984d

web-portal/components/DatePicker.vue
chriswininger: removed

web-portal/components/DateTimePicker.vue
chriswininger: fixed 8b00e02e2bba94b34d4235264f7215cdb46a9131

web-portal/components/DatePicker.vue
chriswininger: https://github.com/infinite-industries/infinite/commit/26c88da4454a1de00277cade533e92664483984d

web-portal/components/DatePicker.vue
chriswininger: fixed 81e13fb2d50ddf4fc16c71420a6756856c629601

MatthewGidcomb (approved):

Issue #491, opened 2024-03-24
Make sure duplicate event detection runs on form init


chriswininger:
This will help Dima detect duplicates when approving. Even we ourselves are often not logged in when submitting

Issue #489, opened 2024-03-03
Editing existing event's times can falsely trigger detection 


MatthewGidcomb:
Given an event with more than one datetime, if you edit one of its datetimes in the admin interface, the existing event detection will flag the event as a duplicate of itself. The endpoint may need to be made aware of whether it's being called for a new or existing event, possibly by passing the event's ID when applicable, so it can be excluded, or the frontend could check the results against the ID of the event.

PR #493, opened 2024-04-11
Duplicate Detection For Event Editing -- issue #491, issue #489

chriswininger:
So that duplicates can be discovered when approving an event we would like to enable it durring issue editing. Three bugs were getting in the way.

1. The duplicate detection was not being triggered on initial form load
2. The state used to determine if the user was an admin was not available on the initial form load
3. The duplicate detection api did not have a way to exclude issues, such as the one you are currently viewing

This pr corrects the two bugs and adds the ability to exclude the issue being edited from detection


MatthewGidcomb (approved):

MatthewGidcomb:
Let's add a Cypress E2E test for this at some point.

PR #496, opened 2024-04-21
Inhouse Pagination

chriswininger:


web-portal/pages/admin-venue-edit/index.vue
```diff
@@ -57,11 +63,50 @@
       }
     },
 
+    mounted: function () {
+      this.$store.dispatch('admin/LoadUnverifiedEvents', { idToken: getToken(this.$auth) })
+      this.$store.dispatch('admin/LoadCurrentEvents', { idToken: getToken(this.$auth) })
+      this.$store.dispatch('admin/LoadResourceEvents', { idToken: getToken(this.$auth) })
+
```
MatthewGidcomb: Is this a copy-paste mistake? I don't think this page needs to fetch all this event data.

MatthewGidcomb:
We can discuss further but I think there's an argument for separating the Vuetify `Pagination` replacement component from the existing portal `Pagination` component, which I should have both named and documented better. The intention there was to simplify displaying data paginated client-side, with either future enhancements or a separate component for server-side pagination. In the latter case, we'll still need a component for managing the various page buttons.

Also, now that I see it, managing the number of page buttons from the components consuming the pagination component seems like the wrong way to handle it. Maybe instead we should consider a hint prop that gives the component an idea of how much space it has to work with (e.g. something like a "size" enum or just a `narrow` bool), and it uses that in combination with the page size to decide where to truncate?

web-portal/pages/admin-venue-edit/index.vue
chriswininger: this was existing code I believe not sure why it's showing in the pr as new, let me look at that

chriswininger: oh nvm, yeah that must be copy/paste mess up, fixed

chriswininger:
> We can discuss further but I think there's an argument for separating the Vuetify `Pagination` replacement component from the existing portal `Pagination` component, which I should have both named and documented better. The intention there was to simplify displaying data paginated client-side, with either future enhancements or a separate component for server-side pagination. In the latter case, we'll still need a component for managing the various page buttons.
> 
> Also, now that I see it, managing the number of page buttons from the components consuming the pagination component seems like the wrong way to handle it. Maybe instead we should consider a hint prop that gives the component an idea of how much space it has to work with (e.g. something like a "size" enum or just a `narrow` bool), and it uses that in combination with the page size to decide where to truncate?

I broke the logic up a bit more so that the part that draws the pagination buttons is separated out from the part that renders the selected pages.

I still think we should stick with letting the pages manager the buttons for now. Honestly long term it's hard to say what the right strategy is, but I was able to make each of those pages function a little better at smaller screens than they did before with this.

If we start using pagination in more places we can re-think. There's definitely still the possibility we ditch this in favor of a third part component in the future, if this proves hard to manage.

MatthewGidcomb (approved):

PR #498, opened 2024-05-23
Updated Edit button on Admin Event List

chriswininger:
Updated the Edit button on the admin event list page to no longer use vuetify

I also added localVue.use(Vuetify) line to the SubmissionForm spec so that several warnings would go away. This does still call one warning about multiple versions of vue being used. This seems like a bug in our vuetify version, but hey it's going away soon. One warning seems better than however many were happing before.

I also removed localVue.use(Vuietify) from the AdminEventList.spec.js file where it's no longer needed. This kills another warning about multiple vue versions

I added a stub to Logo.spec.js to get rid of another test warning

### DIFF
--- a/web-portal/components/AdminEventsList.vue
+++ b/web-portal/components/AdminEventsList.vue
@@ -22,6 +22,7 @@
           <ii-form-button
             style-type="light"
             @click="onEditClicked(calendar_event.id)"
+            :test-id="'edit-issue-button-' + calendar_event.id"
           >
             Edit
           </ii-form-button>
--- a/web-portal/components/AdminEventsList.vue
+++ b/web-portal/components/AdminEventsList.vue
@@ -18,7 +18,14 @@
           <small v-if="calendar_event.date_times.length > 2">(and {{ calendar_event.date_times.length - 2 }} more)</small>
           <template v-if="calendar_event.category && calendar_event.category==='online-resource'">Online Resource</template>
         </td>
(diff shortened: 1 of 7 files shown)


web-portal/components/AdminEventsList.vue
```diff
@@ -18,7 +18,14 @@
...
         </td>
-        <td><v-btn nuxt :to="{ name: 'admin-event-edit-id', params: { id: calendar_event.id } }">Edit</v-btn></td>
+        <td>
+          <ii-form-button
+            style-type="light"
+            @click="onEditClicked(calendar_event.id)"
+          >
+            Edit
+          </ii-form-button>
+        </td>
```
MatthewGidcomb: I think the Cypress tests are failing because this `v-btn` was actually rendering as a link (because of the `nuxt` and `:to` props).

chriswininger: yeah makes sense, will have a look

chriswininger: I debated trying to implement a link version of this but didn't really seem worth it

chriswininger: pushed up a change that should fix this. I went ahead and added a test-id to preserve the spirit of one test which queries for a specific event. I think there was only one even on the page, so I could have simplified it to just check for button, but I like the specificity in that case

MatthewGidcomb: At some point (but it can wait until we rework this for a "design system") I want to revisit this and support rendering as an `a`.

MatthewGidcomb (approved):

PR #503, opened 2024-06-16
Issue #499: Attempt To Supply GPS Coordinates for Venues

chriswininger:
When a new venue is created or a venue is updates, if coordinates are not provided by the request we will attempt to determine them from a GPS lookup service based on the address

### DIFF
--- a/api-server/src/venues/dto/create-update-venue-request.ts
+++ b/api-server/src/venues/dto/create-update-venue-request.ts
@@ -1,6 +1,8 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { IsNotEmpty, IsOptional } from 'class-validator';
 import { Column } from 'sequelize-typescript';
+import isNotNullOrUndefined from '../../utils/is-not-null-or-undefined';
+import cloneAttributes from '../../utils/clone-attributes';
 
 export class CreateVenueRequest {
   @ApiProperty({ example: "Bob Vance's Chill Bar" })
@@ -49,6 +51,12 @@ export class CreateVenueRequest {
   @ApiProperty({ example: 'Downtown' })
   @IsOptional()
   neighborhood?: string;
+
+  constructor(copy?: Partial<CreateVenueRequest>) {
+    if (isNotNullOrUndefined(copy)) {
+      cloneAttributes<CreateVenueRequest>(copy, this);
+    }
+  }
 }
 
 export class UpdateVenueRequest {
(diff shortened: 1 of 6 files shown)


api-server/src/venues/venues.controller.ts
```diff
@@ -81,15 +86,18 @@ export class VenuesController {
...
       .then((venue) => {
-        const venueData = JSON.stringify((venue as any).dataValues, null, 4);
-        this.slackNotificationService.sendNotification(
-          VENUE_SUBMIT,
-          `(${ENV}) New venue created:\n${venueData}`,
+        this.logger.log(
+          `created a new venue -> id: ${venue.id}, name: ${venue.name}`,
         );
+
+        this.sendVenueCreationNotification(venue);
```
chriswininger, line 99: This was a bit of boyscouting. I just created a method for

```
const venueData = JSON.stringify((venue as any).dataValues, null, 4);
        this.slackNotificationService.sendNotification(
          VENUE_SUBMIT,
          `(${ENV}) New venue created:\n${venueData}`)

To make the code easier to read

File: api-server/src/venues/venues.service.ts [hunk 2]
```diff
@@ -37,22 +45,26 @@ export class VenuesService {
...
+  ): Promise<VenueModel> {
+    updatedValues = await this.fillInGPSCoordinatesIfNeededWhenPossible(
+      updatedValues,
+    );
 
-      values = { ...updatedValues, slug };
-    } else {
-      values = { ...updatedValues };
-    }
+    const values = isNotNullOrUndefined(updatedValues.name)
```
chriswininger, line 66: also a bit of scouting, using the ternary ? operator lets this be a bit more concise and allows values to be a constant

api-server/src/venues/dto/create-update-venue-request.ts
```diff
@@ -49,6 +51,12 @@ export class CreateVenueRequest {
   @ApiProperty({ example: 'Downtown' })
   @IsOptional()
   neighborhood?: string;
+
+  constructor(copy?: Partial<CreateVenueRequest>) {
+    if (isNotNullOrUndefined(copy)) {
+      cloneAttributes<CreateVenueRequest>(copy, this);
+    }
+  }
```
MatthewGidcomb, line 59: Not following why this is necessary

chriswininger: I didn't want to mutate state on the values passed into fillInGPSCoordinatesIfNeededWhenPossible. Unfortunately because NestJs makes you use classes rather than `types` you can't simply shallow clone an object with the spread operator.

This is something I've done in several places on the backend to basically give me a away to clone an object, while maintaining it's class.

Honestly I hate it 🤣 but I haven't come up with a better way

chriswininger: I wonder if some of the newly introduced clone methods like https://developer.mozilla.org/en-US/docs/Web/API/structuredClone, might help, haven't had time to dig back into this for a bit 🤷

MatthewGidcomb: I had assumed there was a built-in way to do this.

MatthewGidcomb (approved):

Issue #453, opened 2023-10-22
Add tests for JSON-LD generation


MatthewGidcomb:
We've seen measurable benefits from including JSON-LD metadata in event pages, so it would be wise to add some testing for that functionality.

### Unit Testing

- The logic to produce the metadata content is encapsulated in a static class service for which we could easily write straightforward given-this-event-structure-assert-this-JSON-LD-output unit tests. Low effort, high benefit.
- Testing that the event page includes the metadata may be trickier -- I'm not sure how to tap into the mechanism that updates the document `head` in unit testing. Unknown effort, medium to high benefit.

### E2E Testing

- Add a Cypress test, or modify an existing event creation/editing test, to verify that the appropriate JSON-LD metadata is present in the document `head` on page render. Medium effort, medium to high benefit.

MatthewGidcomb:
Next step: @MatthewGidcomb will scaffold a test suite, with at least one test showing how to set it up, with notes on other test cases.

MatthewGidcomb:
Test suite is scaffolded; ready for someone to finish out for their first contribution.

PR #505, opened 2024-06-23
First pass at tests for JSON-LD metadata service

MatthewGidcomb:
Creates partial unit test suite, room for improvement (good first issue!)

Also adds a proposal for an E2E test that might be useful.

references #453


Digi-D (commented):
Good on my side. This is very detailed.

PR #509, opened 2024-08-12
Integregate anthropic tag prediction

chriswininger:


api-server/src/summarization/summarization.service.ts
```diff
@@ -0,0 +1,110 @@
+import { HttpException, Injectable } from '@nestjs/common';
+import Anthropic from '@anthropic-ai/sdk';
+
+@Injectable()
+export class SummarizationService {
+  private readonly client: Anthropic;
+
+  constructor() {
+    this.client = new Anthropic({ apiKey: process.env['ANTHROPIC_API_KEY'] });
+  }
```
MatthewGidcomb: Can/should we make this fail gracefully if the API key isn't provided?

Maybe it's enough for the portal to gracefully handle errors if the key isn't there.

api-server/src/summarization/summarization.service.ts
```diff
@@ -0,0 +1,110 @@
...
+      throw new BadAnthropicResponse();
+    }
+  }
+}
+
+class BadAnthropicResponse extends Error {
+  constructor() {
+    super('Anthropic Returned A Bad Response');
+  }
+}
```
MatthewGidcomb: If this error will bubble up to the controller, maybe the message shouldn't contain the provider name?

api-server/src/summarization/summarization.service.ts
chriswininger: yeah we should definitely log out some kind of warning and just skip this feature. I want this to all work with or without this enabled

api-server/src/summarization/summarization.service.ts
chriswininger: it shouldn't bubble up to the client this is all logging for server side debugging, this will get swallowed, then if more than the max allowed happens we surface this:


    throw new HttpException(
      'Unable to get a valid response from third party tag provider',
      500,
    );

chriswininger: I think there's a better status code than 500, I was struggling to remember, but seems like there's one for specific third party integration failure 🤷

MatthewGidcomb: Maybe 503 Service Unavailable? It's more for a "maintenance mode" or load issue, but is still somewhat descriptive of what the underlying problem here is.

api-server/src/summarization/summarization.service.ts
chriswininger: added a warning message which will get logged on server start similar to other messages 

image

In this instance I'm updated so that after that initial message it will no longer warn so as not to be overly chatty and if you ask for tags it just returns a empty list

chriswininger: e5000348df6ad16050cd2a0d1359946b518b6036

api-server/src/summarization/summarization.service.ts
chriswininger: sure sounds good

chriswininger: 279a21419778f91622c1c331cb8b008d5a417f75

api-server/src/summarization/summarization.service.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    Output the tags as a JSON array of strings. Do not include any additional text or formatting."""
+    `;
+
+    const errors = [];
+    const numTries = 3;
+    for (let i = 0; i < numTries; i++) {
+      try {
+        if (numTries > 0) {
+          console.info('trying to get a valid message from anthropic again');
```
MatthewGidcomb: Do these logging statements need to use `this.logger`?

chriswininger: good catch, fixed

MatthewGidcomb (approved):

PR #511, opened 2024-09-01
Integrate tag suggestion into submission form

MatthewGidcomb:
Triggered on "description" blur only if tags are currently empty

### DIFF
--- a/web-portal/components/SubmissionForm.vue
+++ b/web-portal/components/SubmissionForm.vue
@@ -626,9 +626,12 @@
         if (this.calendar_event.description && this.calendar_event.tags.length === 0) {
           this.loadingSuggestedTags = true
           this.$tagSuggestionService.getSuggestionsForDescription(this.calendar_event.description)
-            .then(({ data }) => {
-              this.showingSuggestedTags = true
-              this.calendar_event.tags = data
+            .then((suggestions) => {
+              if (suggestions) {
+                this.showingSuggestedTags = true
+                this.calendar_event.tags = suggestions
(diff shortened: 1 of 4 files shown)


web-portal/components/SubmissionForm.vue
```diff
@@ -658,17 +677,7 @@
       },
 
       suggestedTags: function () {
-        return [
```
chriswininger, line 661: I'm confused by this? I thought we would still want to offer auto-complete tags. Rather than using these tags for auto-complete, I thought the idea was we actually populate the form with them.

I guess there's a language challenge here :-) 

To me these are a different kind of "suggested", maybe it's suggested tags vs predicted tags, or autoCompleteTags vs suggestedTags

MatthewGidcomb: I just moved this list to the `TagSuggestionService` rather than have them hard-coded here.

Originally we were going to replace this hard-coded list of "suggested" (maybe "recommended" is more descriptive now?) with a call to the server to get the top X most frequently-used tags, and maybe filter that as you typed in the tag field, but never got around to it.

MatthewGidcomb: Would it be more clear if it was named something like `$tagSuggestionService.getBaseTagSet()`?

chriswininger: oh I see, mostly I was just confused :-)

yeah that works

chriswininger (approved):
+1

PR #512, opened 2024-09-08
Add anthropic key

chriswininger:
this addes the anthropic key

once deployed along with frontend changes the tag suggestion feature
will go live


MatthewGidcomb:
Is the base branch based on `master` ? Lots very old merge commits in the history.

PR #513, opened 2024-09-12
Add analytics gathering for tag suggestion feedback

MatthewGidcomb:
Adds new "internal API" endpoint on web portal to receive structured analytics data on suggested tags. When the AI tag suggestor is used, this endpoint will be POSTed to with the suggested tags, the final tags, and the ID of the event. Data is sent to Slack for later analysis


web-portal/internal-api/slack-analytics.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+
+    // slack-notify doesn't handle this gracefully
+    // hopefully it isn't necessary in production but in dev it's useful
+    if (!SLACK_WEBHOOK_ANALYTICS) {
+      return reject(new Error('No Slack URL configured'))
+    }
+
+    contactChannel.send({
+      channel: 'analytics-test',
```
MatthewGidcomb: Replace with `'analytics'` or other better name before merging

web-portal/components/SubmissionForm.vue
```diff
@@ -636,8 +641,21 @@
             .catch(err => console.error(err))
             .finally(() => {
               this.loadingSuggestedTags = false
+              if (!this.rawSuggestedTags) {
```
chriswininger: I don't think you meant to leave this

chriswininger (approved):
looks good

MatthewGidcomb:
Added environment to the Slack message so we can easily filter out messages from local and staging.

chriswininger (approved):

PR #514, opened 2024-09-22
init: have systemd use 'down' rather than 'stop'

jswank:
Use `docker-compose stop` rather than `docker-compose down`

### DIFF
--- a/ansible/systemd/infinite.service
+++ b/ansible/systemd/infinite.service
@@ -9,7 +9,7 @@ Group=infinite
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files up --detach"
-ExecStop=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files stop"
+ExecStop=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files down"
 
 [Install]
 WantedBy=multi-user.target
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -25,12 +25,6 @@ auth_use_test_users: false
 # The environment {dev|staging|prod}
 env: dev
 
-# The public location of the frontend service (for example https://infinite.industries)
-app_url: fe_url
-
-# The URL of the API
-api_url: api_url
-
 # Timezone handling
 timezone_default: US/Eastern
 timezone_options: US/Eastern,US/Central,US/Mountain,US/Pacific


chriswininger (approved):
awesome, makes sense. I'll try this out in staging

chriswininger:
@jswank I tried this out this morning, unfortunately I'm still seeing the same error

image

ansible/systemd/infinite.service
```diff
@@ -9,7 +9,7 @@ Group=infinite
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files up --detach"
-ExecStop=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files stop"
+ExecStop=/bin/bash -c "docker-compose --project-directory /home/infinite/docker-files down"
```
chriswininger, line 12: I'm confused by this file, does it run up before it runs down?

The process I followed when doing this manually was to:

1. docker-compose pull
2. docker-compose down
3. docker-compose up --detach

I assume the pull is happening somewhere else, or nothing would be working, and maybe I'm not understanding something about this file works, I do not spend a lot of time writing systemd service files

if we're calling down after we call up though, that definitely isn't right

chriswininger, line 12: So I ran this and unfortunately saw the same error, tried a couple of times, still kept seeing the same issue, however, after I went on the box and manually docker-compose down, to kill the only dangling container that was still running from before it started working and has succeeded several times since 🤦

chriswininger: just to confirm:

```
  - name: enable systemd init for docker-compose
    become: true
    systemd:
      name: infinite
      enabled: true
      daemon_reload: true
```

This is what restarts the systemd service?

Does restart invoke both ExecStop and ExecStart and does it wait for the stop command to completely run before the start finishes?

ansible/systemd/infinite.service
chriswininger, line 1: on another note, after successfully deploying the site is up but pretty broken, losts of cors errors that make me think environment variables are not set correctly 

image

chriswininger: There's also a lot of errors causes by our certs having expired, not sure we need valid certs for staging but also not sure I remember this issue from the past

ansible/systemd/infinite.service
jswank: That action enabled installs the systemd unit file.

In the deployment playbook, other actions "notify" the action that restarts docker containers.

ansible/systemd/infinite.service
chriswininger: I think the cert issue may be causing a lot of the problems, maybe the making it so the web-portal is having trouble talking to the api-server

```
web-portal_1   |
web-portal_1   |  ERROR  Cannot read properties of undefined (reading 'Error: certificate has expired')
web-portal_1   |
web-portal_1   |   at setQueryStateFail (helpers/venue-store-helpers/queryState.js:32:0)
web-portal_1   |   at Store.ACTIVE_VENUES_FETCH_FAIL (store/venues.js:49:0)
web-portal_1   |   at wrappedMutationHandler (node_modules/vuex/dist/vuex.common.js:846:13)
web-portal_1   |   at commitIterator (node_modules/vuex/dist/vuex.common.js:468:7)
web-portal_1   |   at Array.forEach (<anonymous>)
web-portal_1   |   at node_modules/vuex/dist/vuex.common.js:467:11
web-portal_1   |   at Store._withCommit (node_modules/vuex/dist/vuex.common.js:626:3)
web-portal_1   |   at Store.commit (node_modules/vuex/dist/vuex.common.js:466:8)
web-portal_1   |   at Store.boundCommit [as commit] (node_modules/vuex/dist/vuex.common.js:411:19)
web-portal_1   |   at local.commit (node_modules/vuex/dist/vuex.common.js:798:13)
web-portal_1   |   at server.js:1233:15
web-portal_1   |   at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
web-portal_1   |
```

chriswininger (commented):
I made one small change here 0365289c1a96bddee309e6273df399a756e1f811, which if nothing else reduces confusing, but I think the big remaining issue is with our certs for staging, maybe related to the fact that we shut down the server.

I don't remember expired certs being an issue in the past though

ansible/systemd/infinite.service
chriswininger: @MatthewGidcomb looks like we get back an error string for a venuid then try to update the state by looking up that error string

```
export function setQueryStateFail(queryState, venueId, error) {
  queryState.queriesByVenueId[venueId].isFetching = false
  queryState.queriesByVenueId[venueId].isSuccess = false
  queryState.queriesByVenueId[venueId].error = error
}
```

`queryState.queriesByVenueId[venueId].isFetching = false`

That's not the root issue, the real problem is that we are bombing on all our api requests, but seems like we do have some bad error handling somewhere on the frontend

chriswininger: also, 😅 I renewed the certs manually (which helpfully we documented in our README) and staging is now up and running 🥳

PR #521, opened 2025-01-12
Error handling enhancement for analytics

chriswininger:
handle analytics failures more correctly

* a failure of the client to sumbit generated vs selected tags should not stop the form
  from showing success on submission of an event
* the web-portal internal api should not return 500 if web hooks are enabled, it should simply
  log a warning to the conolse and return 200

### DIFF
--- a/web-portal/components/SubmissionForm.vue
+++ b/web-portal/components/SubmissionForm.vue
@@ -526,36 +526,33 @@
           reviewed_by_org: this.reviewOrg ? this.reviewOrg : null
         }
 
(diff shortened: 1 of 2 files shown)


MatthewGidcomb:
Since you converted `UploadEvent` to an async function, I wonder if it would be worth it to replace the promise chains with a series of `await`s. I could see it being more readable.

chriswininger:
> Since you converted `UploadEvent` to an async function, I wonder if it would be worth it to replace the promise chains with a series of `await`s. I could see it being more readable.

personally I kind of like the mix-match, but not a strong opinion on this one or they other 🤷

MatthewGidcomb (approved):

PR #522, opened 2025-01-19
Removes the remaining filters

chriswininger:


web-portal/components/admin-venue-edit/AdminEventsListRow.vue
MatthewGidcomb, line 1: Nitpick but can we move this to just `components/` or `components/admin-event-list/`? Seems wrong to put it with the venue components.

web-portal/components/admin-venue-edit/AdminEventsListRow.vue
```diff
@@ -0,0 +1,94 @@
...
+  import FormButton from '@/components/FormButton.vue'
+  import momenttz from 'moment-timezone'
+  import PartnerService from '@/services/PartnerService'
+
+  export default {
+    name: 'AdminEventsListRow',
+    components: {
+      'ii-form-button': FormButton
+    },
+    props: ['calendar_event', 'isProblemUpdated'],
```
MatthewGidcomb: We use `calendar_event` for prop name elsewhere, but it breaks Vue's style guide and I'd like to move away from it at some point. Thoughts on either `calendarEvent` (which would be `calendar-event="source"` in the template) or just `event`?

web-portal/components/admin-venue-edit/AdminEventsListRow.vue
```diff
@@ -0,0 +1,94 @@
...
+    <td>{{ calendar_event.title }}</td>
+
+    <td>
+      <div v-for="(date_time, index) in calendar_event.date_times.slice(0, 2)" :key="index">
+        {{ dateFormat(date_time) }}
+      </div>
+
+      <small v-if="calendar_event.date_times.length > 2">
+        (and {{ calendar_event.date_times.length - 2 }} more)
+      </small>
```
MatthewGidcomb: If we wanted to handle this with a computed property, we could have it return an object with one or two pre-formatted dates + the count of how many more there are, something like:

```js
when: function () {
  return {
    formattedDates: calendar_event.date_times?.slice(0, 2).map((dt) => this.dateFormat(dt)),
    more: calendar_event.date_times?.length > 2 ? calendar_event.date_times.length - 2 : 0
  }
}
```

And then the template could use that like this:

```vue
      <div v-for="(formattedDate, index) in when.formattedDates" :key="index">
        {{ formattedDate }}
      </div>

      <small v-if="when.more">
        (and {{ when.more }} more)
      </small>
```

web-portal/components/AdminEventsList.vue
```diff
@@ -59,41 +38,23 @@
...
-      dateFormat: function (date) {
-        return [
-          momenttz(date.start_time).tz(date.timezone).format('dddd, MMMM Do h:mma'),
-          '-',
-          momenttz(date.end_time).tz(date.timezone).format('h:mma z')
-        ].join(' ')
-      },
       ownerLogo: function (owner) {
         return PartnerService.getLogoForReviewer(owner)
       }
```
MatthewGidcomb: Will need to delete this one as well.

web-portal/components/admin-venue-edit/AdminEventsListRow.vue
chriswininger: :heavy_check_mark:

web-portal/components/admin-venue-edit/AdminEventsListRow.vue
chriswininger: :heavy_check_mark:

web-portal/components/admin-venue-edit/AdminEventsListRow.vue
chriswininger: thx, probably just leave it alone for now

web-portal/components/AdminEventsList.vue
chriswininger: done :heavy_check_mark:

MatthewGidcomb (approved):

PR #526, opened 2025-03-07
Upgrade web portal to Nuxt 3

### LINKED ISSUE #407: Investigate upgrading to Nuxt (and Vue) 3
The web portal is built with Nuxt, and Vue, 2. Vue 2 end-of-life has been set for December 31, 2023 ([more](https://v2.vuejs.org/lts/)), and Nuxt has decided to do the same for [Nuxt 2](https://nuxt.com/blog/vision-2023#migrating-to-nuxt-3).

This isn't so urgent that we need to drop everything and start upgrading -- I think we'd be fine to coast along on Nuxt+Vue 2 for a while, probably -- but definitely something to consider. Upgrade likely won't be trivial but shouldn't be nearly as hard as the past three rewrites.

MatthewGidcomb:
Closes #407

[Tracking project](https://github.com/orgs/infinite-industries/projects/5)


chriswininger (approved):
looks great kid, don't worry about that definitely not written by an LLM isEqual function

PR #535, opened 2025-09-14
Add date range filtering all verified endpoint

chriswininger:
Allow for optional filtering by date range

The verified endpoint now accepts an optional start and end date

you must provide both

when present we will only return events within that range

pagination will still apply


api-server/src/events/events.controller.ts
```diff
@@ -128,10 +128,34 @@ export class EventsController {
...
+    type: String,
+  })
+  @ApiImplicitQuery({
+    name: 'endDate',
+    description:
+      'End date for filtering events (ISO 8601 format). Must be provided together with startDate.',
+    example: '2024-12-31T23:59:59.999Z',
+    required: false,
+    type: String,
+  })
```
MatthewGidcomb: If we always need to include both, what would you think of doing one `dateRange` param with both dates separated by `/` (the [interval](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) format)?

```dateRange=2025-01-01T00:00:00Z/2025-06-30T23:59:59Z```

chriswininger: :thinking: I haven't seen that used much so it makes me a bit nervous, but I was stumbling over that a bit too. It would be simpler if this were a post request, we'd just do json. We could do json in the query param too though i don't know how common that is.

if we went with the string separator approach we'd still have to have parse it and verify they sent both because the client could still leave one out though it does make the intent clearer

chriswininger: the more I think on it the more I like it though I think I'll make the change

MatthewGidcomb (approved):

PR #537, opened 2025-09-21
Implement Partner Work Flows on API - Part I

chriswininger:
1. Created migration to add partner and user/partner membership tables
2. Created migration to add owning_partner_id to the event table
3. Created related sequalize models and updated existing models and dtos to interact with new tables
4. created middle-ware for consolidating the resolution of user information on each request
5. updated guards to leverage the new middle-ware and created new guards to deal with new authenticated situations such as an authenticated user that is not an infinite-admin but may be a partner-admin
6. fixed bug in filtering of organizer contact e-mail that was leaving it on the event for some un-authenticated requests
7. updated filtering logic to leave on organizer-contact and fetch event admin metadata if you are a partner-admin that owns the event via partner association
8. created some basic crud endpoints around partners and partner associations
9. created a new endpoint for fetching all unverified events that belong to you via partner-association (if you are an infinite-admin they all belong to you and it falls through to existing get all unverified logic)

### DIFF
--- a/api-server/migrations/20250921214442-create-users-partners-mappings.js
+++ b/api-server/migrations/20250921214442-create-users-partners-mappings.js
@@ -1,0 +1,67 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.createTable('users_partners_mappings', {
+      id: {
+        allowNull: false,
+        primaryKey: true,
+        type: Sequelize.UUID,
+        defaultValue: Sequelize.literal('uuid_generate_v4()')
+      },
+      user_id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: 'users',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
+      },
+      partner_id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: 'partners',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
+      },
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.literal('now()')
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.literal('now()')
+      }
+    });
+
+    // Add a composite unique constraint to prevent duplicate user-partner mappings
+    await queryInterface.addConstraint('users_partners_mappings', {
+      fields: ['user_id', 'partner_id'],
+      type: 'unique',
+      name: 'unique_user_partner_mapping'
+    });
+
+    // Add indexes for better query performance
+    await queryInterface.addIndex('users_partners_mappings', {
+      fields: ['user_id'],
+      name: 'idx_users_partners_mappings_user_id'
+    });
+
+    await queryInterface.addIndex('users_partners_mappings', {
+      fields: ['partner_id'],
+      name: 'idx_users_partners_mappings_partner_id'
+    });
+  },
+
+  async down (queryInterface, Sequelize) {
+    await queryInterface.dropTable('users_partners_mappings');
+  }
+};
--- a/api-server/src/announcements/announcements.controller.ts
+++ b/api-server/src/announcements/announcements.controller.ts
@@ -9,7 +9,7 @@ import {
 } from '@nestjs/common';
 import { VERSION_1_URI } from '../utils/versionts';
 import { ApiOperation, ApiTags } from '@nestjs/swagger';
-import { AdminAuthGuard } from '../authentication/AdminAuth.guard';
+import { AdminAuthGuard } from '../authentication/auth-guards/admin-auth.guard';
 import { AnnouncementsService } from './announcements.service';
 import { CreateOrUpdateAnnouncementRequest } from './dto/create-or-update-announcement-request';
 import { AnnouncementResponse } from './dto/announcement-response';
--- a/api-server/src/announcements/announcements.controller.ts
+++ b/api-server/src/announcements/announcements.controller.ts
@@ -9,7 +9,7 @@ import {
 } from '@nestjs/common';
 import { VERSION_1_URI } from '../utils/versionts';
 import { ApiOperation, ApiTags } from '@nestjs/swagger';
-import { AuthGuard } from '../authentication/auth.guard';
+import { AdminAuthGuard } from '../authentication/AdminAuth.guard';
 import { AnnouncementsService } from './announcements.service';
 import { CreateOrUpdateAnnouncementRequest } from './dto/create-or-update-announcement-request';
 import { AnnouncementResponse } from './dto/announcement-response';
@@ -31,7 +31,7 @@ export class AnnouncementsController {
   }
 
   @Post('/ensure-one-announcement')
-  @UseGuards(AuthGuard)
+  @UseGuards(AdminAuthGuard)
   @ApiOperation({
     summary:
       'Create an announcement if it does not exist or update the current one if present.',
@@ -48,7 +48,7 @@ export class AnnouncementsController {
   }
 
   @Put('/:id')
-  @UseGuards(AuthGuard)
+  @UseGuards(AdminAuthGuard)
   @ApiOperation({ summary: 'update an existing announcement' })
   updateAnnouncement(
     @Param() params: FindByIdParams,
--- a/api-server/src/app.module.ts
+++ b/api-server/src/app.module.ts
@@ -91,7 +91,9 @@ const dialectOptions = SQL_IS_USING_SSL
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
     consumer
-      .apply(UserInformationMiddleware, LoggingMiddleware) // both apply to all routes
+      // for all routes, including root, attach user information to the request
+      // and run request logging middleware
+      .apply(UserInformationMiddleware, LoggingMiddleware)
       .forRoutes('*');
   }
 }
--- a/api-server/src/app.module.ts
+++ b/api-server/src/app.module.ts
@@ -91,9 +91,7 @@ const dialectOptions = SQL_IS_USING_SSL
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
     consumer
-      .apply(UserInformationMiddleware) // attaches userInformation to the Request object
-      .forRoutes('*')
-      .apply(LoggingMiddleware)
-      .forRoutes('/**');
+      .apply(UserInformationMiddleware, LoggingMiddleware) // both apply to all routes
+      .forRoutes('*');
   }
 }
--- a/api-server/src/app.module.ts
+++ b/api-server/src/app.module.ts
@@ -6,6 +6,7 @@ import { AnnouncementsModule } from './announcements/announcements.module';
 import { UsersModules } from './users/users.modules';
 import { NotificationsModule } from './notifications/notifications.module';
 import LoggingMiddleware from './logging/logging.middleware';
+import { UserInformationMiddleware } from './authentication/user-information.middleware';
 import { WinstonModule } from 'nest-winston';
 import { format, transports } from 'winston';
 import { CalendaringModule } from './calendaring/calendaring.module';
@@ -84,10 +85,15 @@ const dialectOptions = SQL_IS_USING_SSL
       provide: APP_FILTER,
       useClass: ExceptionLogger,
     },
+    UserInformationMiddleware,
   ],
 })
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
-    consumer.apply(LoggingMiddleware).forRoutes('/**');
+    consumer
+      .apply(UserInformationMiddleware) // attaches userInformation to the Request object
+      .forRoutes('*')
+      .apply(LoggingMiddleware)
+      .forRoutes('/**');
   }
 }
--- a/api-server/src/authentication/auth-guards/admin-auth.guard.ts
+++ b/api-server/src/authentication/auth-guards/admin-auth.guard.ts
@@ -1,5 +1,5 @@
 import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
-import { RequestWithUserInfo } from '../users/dto/RequestWithUserInfo';
+import { RequestWithUserInfo } from '../../users/dto/RequestWithUserInfo';
 
 @Injectable()
 export class AdminAuthGuard implements CanActivate {
--- a/api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
+++ b/api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
@@ -1,5 +1,3 @@
-import { Request } from 'express';
-import isAdminUser from '../is-admin-user';
 import EventDTO from '../../events/dto/eventDTO';
 import { RequestWithUserInfo } from '../../users/dto/RequestWithUserInfo';
 import isNotNullOrUndefined from '../../utils/is-not-null-or-undefined';
--- a/api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
+++ b/api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
@@ -3,6 +3,7 @@ import isAdminUser from '../is-admin-user';
 import EventDTO from '../../events/dto/eventDTO';
 import { RequestWithUserInfo } from '../../users/dto/RequestWithUserInfo';
 import isNotNullOrUndefined from '../../utils/is-not-null-or-undefined';
+import { EventModel } from '../../events/models/event.model';
 
 type EventOrEventList = EventDTO | EventDTO[];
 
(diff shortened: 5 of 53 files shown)


api-server/src/announcements/announcements.controller.ts
```diff
@@ -31,7 +31,7 @@ export class AnnouncementsController {
   }
 
   @Post('/ensure-one-announcement')
-  @UseGuards(AuthGuard)
+  @UseGuards(AdminAuthGuard)
```
chriswininger, line 34: We no have to be more explicit in this controller, all of the endpoints require that you are at least authenticated, but some require full infinite-admin access while others require only that you a partner or an infinite-admin

I added quite a bit more testing around this and other auth scenarios as it was all getting quite confusing, and found some preexisting bugs in the process

api-server/src/app.module.ts
```diff
@@ -84,10 +85,15 @@ const dialectOptions = SQL_IS_USING_SSL
...
+    UserInformationMiddleware,
   ],
 })
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
-    consumer.apply(LoggingMiddleware).forRoutes('/**');
+    consumer
+      // for all routes, including root, attach user information to the request
+      // and run request logging middleware
+      .apply(UserInformationMiddleware, LoggingMiddleware)
```
chriswininger, line 96: This is part of the consolidation of user handling, this middleware happens on all requests, if a token is present it constructs a user information object by combing data from the token and the database entry associated with that users. This logic is not entirely new, but was previously only being done for admin users and only on certain requests that were annotated as requiring auth.

We want to do it on all requests. Even if the endpoint doesn't require auth, you may be authenticated and we want easy access to that user information so we can do things like decide if we should filter out contact e-mails

api-server/src/authentication/auth-guards/admin-auth.guard.ts
```diff
@@ -0,0 +1,11 @@
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
```
chriswininger, line 1: this shows as a new file, but it's not entirely, however it was renamed and vastly simplified and now it has sisters for enforcing partner-admin, or simply authenticated no admin status (that last one could possibly be re-thought, but I put on the authenticated controllers at the top level then put for fine grained guards on the individual endpoints)

api-server/src/authentication/user-information.middleware.ts
chriswininger, line 1: This is the new middleware, most the meet happens in the userService. I'd been trying in the past to not pass the Request object into the db related service to maintain separation of concern but it was really making life harder so I gave up here. It's much nicer to have a service method that takes the request and and always returns a fully formed user object with consistent logic, though in theory we shouldn't need to call this outside of the middlware. Everywhere else should be able grab the user off of the request.

api-server/src/events/dto/eventDTO.ts
```diff
@@ -35,7 +42,12 @@ export default class EventDTO {
   @ApiProperty({ example: '5' })
   admission_fee: string;
 
-  @ApiProperty({ example: 'bob.vance@refridgeration.com' })
+  @ApiProperty({
+    description: `The e-mail address for admins of the site to communicate with
+    about the event. This will be filtered out of un-authenticated api calls. It
+    is meant for communication between the submitter and site admins`,
+    example: 'bob.vance@refridgeration.com',
+  })
```
chriswininger, line 50: noticed we had no description here and backfilled it because I had to remind myself what a lot of this stuff was for :-)

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -40,16 +41,25 @@ import {
...
 import { validateAndExtractOptionalDateTimeFilters } from './utils/validateAndExtractOptionalDatetimeFilters';
+import { AuthenticatedUserGuard } from '../authentication/auth-guards/authenticated-user.guard';
+import { PartnerAdminGuard } from '../authentication/auth-guards/partner-admin.guard';
+import UsersService from '../users/users.service';
+import { RequestWithUserInfo } from '../users/dto/RequestWithUserInfo';
+import { Op } from 'sequelize';
 
 @Controller(`${VERSION_1_URI}/authenticated/events`)
-@UseGuards(AuthGuard)
+@UseGuards(AuthenticatedUserGuard)
```
chriswininger, line 50: In retrospect I think I could have just used the PartnerGuard on all of these and then applied the additional admin guard on the ones that needed, still debating in my head what is more readable, effect would be the same :shrug:

api-server/src/events/events.controller.ts
```diff
@@ -72,13 +71,13 @@ export class EventsController {
...
+    @Req() request: RequestWithUserInfo,
   ): Promise<EventsResponse> {
-    const isAdmin = await isAdminUser(request);
+    const isAdmin = request.userInformation?.isInfiniteAdmin || false;
 
     const include = isAdmin
-      ? [VenueModel, DatetimeVenueModel, EventAdminMetadataModel]
-      : [VenueModel, DatetimeVenueModel];
+      ? [VenueModel, DatetimeVenueModel, EventAdminMetadataModel, PartnerModel]
+      : [VenueModel, DatetimeVenueModel, PartnerModel];
```
chriswininger, line 80: just some plumbing had to do this all over the place since the middelware is happening on all requests. I was trying to figure out if there was a way to say UserModel, VenueModel,etc should just always be available but couldn't quite figure it out. I'm also no 100% sure that all of these are needed in the final version of this pr after I did some more cleanup and consolidation. I may experiment with it some more, but I'm not too worried about it

api-server/src/events/events.controller.ts
```diff
@@ -171,7 +168,7 @@ export class EventsController {
         verifiedOnly: true,
         startDate,
         endDate,
-        isUserAdmin,
+        request,
       })
```
chriswininger, line 172: passing the request now as the logic is now a little more complicated than just isUserAdmin and it felt best to consolidate it in the service layer

api-server/src/events/events.module.ts
```diff
@@ -9,8 +9,11 @@ import { NotificationsModule } from '../notifications/notifications.module';
 import { DatetimeVenueModel } from './models/datetime-venue.model';
 import { EventAdminMetadataModel } from './models/event-admin-metadata.model';
 import { VenueModel } from '../venues/models/venue.model';
+import { PartnerModel } from '../users/models/partner.model';
 import ExistingEventDetectionService from './existing-event-detection-service';
 import { ExistingEventDetectionController } from './existing-event-detection-controller';
+import { UserModel } from '../users/users.modules';
+import UsersService from '../users/users.service';
```
chriswininger, line 16: more of the aforementioned nest dependency injection plumbing madness, that is probably though possible not still needed :-) Again not too worried about it but may take another pass at a way to make this cleaner

api-server/src/users/partners.authenticated.controller.ts
chriswininger, line 1: this is a controller for managing partners, we can do a lot of this direct in the db at first if you want or using the swagger UI. Eventually I'd like to build a UI around it.

Mostly I built it out though to make it easier to test things as I went :-)

api-server/src/users/partners.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    try {
+      return await this.partnersModel.create({
+        id: v4(),
+        name: createPartnerRequest.name,
+        logo_url: createPartnerRequest.logo_url || null,
+      });
+    } catch (error) {
+      // Check if it's a unique constraint violation
+      if (error.name === 'SequelizeUniqueConstraintError') {
+        throw new ConflictException('Partner with this name already exists');
```
chriswininger, line 44: I let the LLM build out some of this and it was way more specific with error handling that I usually would be. I considered dumbing some of it down, but  honestly it made sense and seemed to all work as expected when tested, so I'm going to call this a win for the machines :angry:

api-server/src/users/users.controller.ts
```diff
@@ -30,25 +30,36 @@ export class UsersController {
   async getCurrentUser(
     @Req() request: RequestWithUserInfo,
   ): Promise<UserInfoResp> {
-    const userInfo: UserInformation = request.userInformation;
-    const userInfoToPersist = buildFromUserInfo(userInfo);
+    return await this.userService.ensureCurrentUserByName(request);
```
chriswininger: ah wrote this before I further consolidated this into middleware, I bit this can be even further simplified now :-) will fix

api-server/src/users/users.controller.ts
```diff
@@ -30,25 +30,36 @@ export class UsersController {
...
+  @ApiResponse({
+    status: 200,
+    description: 'List of partners associated with the user',
+    type: PartnersListResponse,
+  })
+  @ApiResponse({
+    status: 401,
+    description: 'Unauthorized - user not authenticated',
+  })
+  async getPartnersForUser(
```
chriswininger, line 50: this endpoint will return the list of partners associated with your user. i think this maybe useful for some UI ideas I had.

api-server/src/users/users.service.ts
chriswininger, line 1: This is the meat and potatoes of user fetch consolidation. Much of this logic, except the parts dealing with partners is not new, but it's been shifted into this service method as I was finding a need to access it more places, though in the end I found that I could make things really simple by invoking it once in middlware, so most code should not need to call this directly

api-server/src/users/users.service.ts
```diff
@@ -17,7 +79,17 @@ export default class UsersService {
...
     return this.usersModel
-      .findOrCreate({ where: { name }, defaults: userToInsert })
+      .findOrCreate({
+        where: { name },
+        defaults: userToInsert,
+        include: [
+          {
+            model: PartnerModel,
+            as: 'partners',
+            through: { attributes: [] }, // Exclude join table attributes, for example owner_partner_id
```
chriswininger, line 90: this was the LLMs idea, I'm not sure if I like it or not. I tend to prefer passing all the things along, though I get why it's redundant since we will return the full partner objects which have the id :shrug:

api-server/test/events.authenticated.e2e-spec.ts
```diff
@@ -718,6 +726,604 @@ describe('Authenticated Events API', () => {
...
+          const eventIds = events.map((event: any) => event.id);
+          expect(eventIds).toContain(partner1Event.id);
+          expect(eventIds).toContain(partner2Event.id);
+          expect(eventIds).toContain(noPartnerEvent.id);
+        });
+    });
+  });
+
+  describe('Non-Admin Access Tests', () => {
+    let nonAdminToken: string;
```
chriswininger, line 1206: sorry for the spam lots of test backfilling, it proved useful as I broke things several times throughout

api-server/test/events.e2e-spec.ts
```diff
@@ -107,7 +116,7 @@ describe('Events API', () => {
           const paginatedEventReturned = events[i];
           const expectedEvent = allEvents[i];
 
-          assertEventsEqual(paginatedEventReturned, expectedEvent);
+          assertEventsEqual(paginatedEventReturned, expectedEvent, false);
```
chriswininger, line 119: this was a fun bug I caught. these tests should have been failing out setting that last flag to false, which tells the assertion whether it would expect the organizer e-mail to be there. We weren't filtering it properly on these endpoints, now we are

api-server/test/events.e2e-spec.ts
```diff
@@ -991,4 +1000,368 @@ describe('Events API', () => {
...
+          const { event: eventReturned, status } = body;
+
+          expect(status).toEqual('success');
+          // Partner admin should NOT see organizer_contact for events they don't own
+          expect(eventReturned.organizer_contact).toBeUndefined();
+        });
+    });
+  });
+
+  async function associateUserWithPartners(
```
chriswininger, line 1354: This has gotten defined in a few spots, it should probably to be pulled out into a shared utility.

api-server/todo.md
chriswininger, line 1: obviously I need to delete this before we merge lol

api-server/migrations/20250921214442-create-users-partners-mappings.js
```diff
@@ -0,0 +1,67 @@
...
+      partner_id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: 'partners',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
+      },
```
MatthewGidcomb, line 32: Might be overkill or overly hard to do with Sequelize, but could the primary key for this table be the (user ID, partner ID) pair?

api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
```diff
@@ -1,37 +1,76 @@
...
     return events;
   }
 
   if (isGenericEventList(events)) {
-    return removeSensitiveDataList(events) as T;
+    return removeSensitiveDataList(request, events) as T;
   } else {
-    return removeSensitiveDataForSingleEvent(events as EventDTO) as T;
+    return removeSensitiveDataForSingleEvent(request, events as EventDTO) as T;
   }
```
MatthewGidcomb, line 22: Something we may  want to discuss: do we actually need to filter sensitive data (organizer contact email) for partner admins? I think probably not, since they're sort of the "owners" of these events it makes sense for them to be able to contact the organizer (if it's an external submission), but I could be talked out of this.

Also, there's the practical/technical concern of what happens if we exclude this but still let partners edit owned events -- does that break validation of the submission form?

api-server/src/authentication/user-information.middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+        request,
+      );
+    } catch (ex) {
+      // We run this for all request authenticated or not, if there was parse
+      // error it was already logged and rethrown by ensureCurrentUserByName,
+      // we can ignore it here
+      this.logger.debug(`User information parsing failed: ${ex.message}`);
+    }
+
+    next();
```
MatthewGidcomb, line 44: This doesn't matter for the way this is currently written, but stylistically and/or for future-proofing I wonder if the `next()` call should be in a `finally` block?

api-server/src/events/dto/eventDTO.ts
MatthewGidcomb: Good to document this -- I remember its purpose based on the description text we have on the submission form...

api-server/src/events/dto/eventDTO.ts
```diff
@@ -97,6 +109,21 @@ export default class EventDTO {
...
+    example: {
+      id: '123e4567-e89b-12d3-a456-426614174000',
+      name: 'TechCorp Inc.',
+      logo_url: 'https://example.com/logo.png',
+      createdAt: '2024-01-15T10:30:00.000Z',
+      updatedAt: '2024-01-15T10:30:00.000Z',
+    },
+  })
+  owning_partner?: PartnerDTO;
+
```
MatthewGidcomb, line 129: What do we do with `reviewed_by_org` -- mark it as deprecated?

api-server/src/events/events.service.ts
```diff
@@ -168,18 +176,48 @@ export class EventsService {
...
+        partnerIds.length > 0
+          ? await this.partnerModel.findAll({
+              where: {
+                id: {
+                  [Op.or]: partnerIds,
+                },
+              },
+            })
+          : [];
+
```
MatthewGidcomb, line 194: Does Sequelize not support fetching the related partners as part of the events query?

api-server/src/users/dto/RequestWithUserInfo.ts
```diff
@@ -0,0 +1,9 @@
+import { Request } from 'express';
+import { UserInfoResp } from './user-info-resp';
+
+// This extends with built-in request interface with type information about
+// the current user. We affix this to the request via the suer-information-middleware
```
MatthewGidcomb:
```suggestion
// the current user. We affix this to the request via the user-information-middleware
```

api-server/src/users/partners.authenticated.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+    description: 'Conflict - partner with this name already exists',
+  })
+  async create(
+    @Body() createPartnerRequest: CreatePartnerRequest,
+    @Req() request: Request,
+  ): Promise<PartnerDTO> {
+    const isAdmin = await isAdminUser(request);
+    if (!isAdmin) {
+      throw new ForbiddenException(FORBIDDEN_ERROR_MESSAGE);
+    }
```
MatthewGidcomb: Is this not handled by `AdminAuthGuard`?

api-server/src/users/partners.service.ts
MatthewGidcomb: At least this specific case of the error handling looks like a good idea to me.

I haven't kept up with Nest like I should, given our dependence on it, but I wonder if its maintainers have made a point of having good examples available to the AI companies?

api-server/migrations/20250921214442-create-users-partners-mappings.js
chriswininger: I don't think so, I could be mixing it up with other orms but pretty sure sequelize is fairly insistent about always having a uuid, and in any case it's not a bad rule, even if you barely use it can come in handy in migration scenarios, but yeah definitely feels extra :-)

api-server/src/authentication/filters/remove-sensitive-data-for-non-admins.ts
chriswininger: exactly we don't want to :-) i'm getting confused at this point at this point about what is in this pr and what is in PR part 2, but how this all should work is that if you are an infinite-admin or a partner-admin that owns a given event then you will get all the data for that event otherwise you will get the trimmed version 

you have to apply this operation to all the entries in the list, at least for a partner-admin because the endpoint could be returning a mix of owned and un-owned events

api-server/src/authentication/user-information.middleware.ts
chriswininger: we could tough won't that be effectively the same since we are catching and not re-throwing all errors?

chriswininger: i guess that's what you mean by it doesn't matter with the way it's written :-)

chriswininger: so the difference would be that if we ever re-through something inside our catch then it would still run next?

api-server/src/events/dto/eventDTO.ts
chriswininger: yeah, i was of thinking of running some sql in prod to just set any thing with an entry currently there to be owned by wrfl and then we could have migration to drop this

chriswininger: f2709101d35b82fa58b837360aa5a6fa01e5efd3

added deprecation flags and comments

api-server/src/events/events.service.ts
chriswininger: It does but non the context of current events. i believe there is a comment somewhere in this file explaining it.

We to get current events to do what current events does with all the weirdness required for multi-day events i had to roll custom sql and it breaks the automatic stuff.

This is how it's been for quite a while sadly, but it's not too bad performance wise because we are doing pagination on most of the endpoinst now and we only fetch the additional resources post pagination

One thing i keep meaning to do though (or maybe i have :rofl:) is to enforce a page-size max on all the endpoints

api-server/src/users/partners.authenticated.controller.ts
chriswininger: you're right it is, I created this one early in the process before I had sorted how all those guards were working :-) I noticed a few other existing spots in the code like this too where I had been confused. I relearned a lot in the process lol

api-server/src/events/dto/eventDTO.ts
chriswininger: whoops wrong branch acade2878c15d59a590fd41119a76b6c07a34d00

api-server/src/users/partners.authenticated.controller.ts
chriswininger: fixed, good catch 07c63479410f19746a353c81e9b165b7ccdfe636

chriswininger: wrote this bit early in the process when I still confused about auth guards :rofl:

api-server/src/users/partners.service.ts
chriswininger: it couldn't hurt (well it could if leaks information useful to an attacker :rofl:). I would have had error handling just would have been more lazy and generic about it, but this is good :-)

api-server/src/authentication/user-information.middleware.ts
MatthewGidcomb: Oh, that's a good point. Not sure what would happen if we re-throw an exception in the `catch` that we want to result in a request with 400/500 error code, and then `next()` in the `finally`.

PR #539, opened 2025-09-30
Migrate submit-event to composition api

chriswininger:
I was experiencing strange behavior trying to add some new network requests to this file for the partner work. All the rough edges seem to follow off once we get rid of the object api stuff.

### DIFF
--- a/web-portal/pages/submit-event.vue
+++ b/web-portal/pages/submit-event.vue
@@ -101,132 +101,135 @@
 
 </template>
 
-<script>
+<script setup>
+  import { ref, onMounted, onBeforeUnmount, nextTick } from 'vue'
   import { useStore } from 'vuex'
+  import { useRoute, onBeforeRouteLeave } from 'vue-router'
   import SubmissionForm from '@/components/SubmissionForm.vue'
   import SubmissionPreview from '@/components/SubmissionPreview'
   import PartnerService from '@/services/PartnerService'
   import { FETCH_ACTIVE_VENUES } from '../store/venues'
 
-  export default {
-    async setup () {
-      definePageMeta({
-        layout: 'no-mobile-cta',
-      })
-      const { query } = useRoute()
-      const store = useStore()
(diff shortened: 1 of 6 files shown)


web-portal/pages/submit-event.vue
```diff
@@ -101,132 +101,135 @@
...
+  })
+
+  const { query } = useRoute()
+  const store = useStore()
+
+  const mode = ref('edit')
+  const previewEvent = ref(null)
+  const partner = ref(null)
+  const submissionErrorMessage = ref(null)
+  const form = ref(null)
```
MatthewGidcomb, line 124: Per [this](https://vuejs.org/guide/essentials/template-refs.html) there's a special helper for template refs:

```suggestion
  const form = useTemplateRef('form')
```

Can't say I like the API change here on Vue's part

chriswininger: unfortunately useTemplateRef seems to break it, I'm not sure why, but if I switch to that, when I hit the final submit button on the preview screen it just goes into an endless spinner, no errors that I can see, it just gets stuck there :shrug: seems to work fine with ref though

chriswininger:
```
  const submitEventForReal = () => {
    previewEvent.value = null
    mode.value = 'pending'
    console.log('!!! much upload for real: ' + form.UploadEvent + ', ' + form.value)
    form.value?.UploadEvent()
  }
  ```
  
  if I use useTemplateRef form.value seems to be null, form.UploadEvent also does not exit

MatthewGidcomb: Interesting, I tried it locally and it worked for both `form` and `previewHeading`. I'm fine with the way you have it, though, seems like this was the pre-Vue 3.5 way (I _really_ don't care for the "magic" inherent in "use the same variable name as the ref in the template")

MatthewGidcomb (approved):

PR #541, opened 2025-10-12
Fix frontend handling of image upload errors

MatthewGidcomb:
This is a quick fix for the issue noted in today's meeting, that errors from image upload problems on the submission form that previously displayed to the user weren't appearing. Looks like we just overlooked that handling during the Nuxt upgrade, which replaced `axios` with `fetch`, which has different response APIs.

While I was looking at this code I made some styling tweaks to the displayed error message that I think will make it stand out more vs the "please ping us" message.


chriswininger (approved):
oh yep, fixed so many cases like this, missed this one :+1:

PR #543, opened 2025-11-16
Partner work flows add partner user relations api part ii

chriswininger:
Partner Work Flog -- Part II

Adds support for partner admins, modifying the api the thread through partner information and implementing access control for partner-admins


.github/workflows/cypress-tests.yml
```diff
@@ -71,12 +71,21 @@ jobs:
...
         run: |
-          echo "{\"admin_auth_username\":\"$ENV_US\",\"admin_auth_password\":\"$ENV_PW\"}" > cypress.env.json
+          cat > cypress.env.json << EOF
+          {
+            "admin_auth_username": "$ENV_US",
+            "admin_auth_password": "$ENV_PW",
+            "partner_admin_username": "$PARTNER_ENV_US",
+            "partner_admin_password": "$PARTNER_ENV_PW"
+          }
+          EOF
```
MatthewGidcomb, line 81: This is much more readable, good call.

api-server/src/authentication/authentication.controller.ts
```diff
@@ -116,4 +142,52 @@ export class AuthenticationController {
...
   }
+
+  private async ensureTestPartnerUser() {
+    this.logger.log('Ensuring partner-admin user exists...');
+
+    // Ensure the user exists
+    const testPartnerUser = await this.usersService.ensureByName({
+      name: 'partner-admin',
+      nickname: 'partner-admin',
+      picture: 'https://via.placeholder.com/150',
```
MatthewGidcomb, line 153: Not a big deal but I wonder if we should consider moving the details of the test partner out into an external file (or top-level-of-this-file constant), so we don't have to worry about syncing this and the `createJWTForTestUser` call above if we change anything?

MatthewGidcomb: (same thought applies to the "Random Displacement Shipping" partner details below)

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -96,10 +103,19 @@ export default class EventsAuthenticatedController {
     type: String,
     description: 'Filter events by category',
   })
+  @ApiQuery({
+    name: 'owning_partner_id',
+    description:
+      'Filter events by owning partner ID. Can be specified multiple times.',
```
MatthewGidcomb: Nitpick: would this be clearer as "More than one partner ID can be specified" or "Supports array syntax to return events for more than one partner at a time"?

api-server/src/events/events.service.ts
```diff
@@ -58,18 +63,40 @@ export class EventsService {
...
       options = { ...findOptions, ...options };
     }
 
-    return this.eventModel.findOne(options).then((result) => {
-      return result;
+    // findOne will add the include option with all the necessary embedded models
+    return await this.findOne(options)
+      .then((result) => {
+        return result;
+      })
```
MatthewGidcomb: Is this step necessary?

api-server/src/events/events.service.ts
```diff
@@ -58,18 +63,40 @@ export class EventsService {
...
+        } else {
+          return event;
+        }
+      })
+      .then((event) => removeSensitiveDataForSingleEvent(request, event));
+  }
+
+  // just a private utility with no auth checks for use internally
+  // it automatically applies includes
+  private async findOne(findOptions: FindOptions): Promise<EventModel> {
```
MatthewGidcomb: What about calling this helper `findOneWithRelated`, to be explicit about its purpose?

web-portal/components/AdminEventsListRow.vue
```diff
@@ -68,7 +68,7 @@
         }
       },
       ownerLogo: function () {
-        return PartnerService.getLogoForReviewer(this.calendarEvent.reviewed_by_org)
+        return this.calendarEvent?.owning_partner?.logo_url
```
MatthewGidcomb, line 71: Glad the partner itself can be embedded; that will make this much easier on the frontend.

api-server/src/events/events.authenticated.controller.ts
chriswininger: f950d54e9e0a2d20351a23e36702c668a0c9a802

api-server/src/events/events.service.ts
chriswininger: 33e1c261e4af48832f3a1bd9849f4b952bab451d

api-server/src/events/events.service.ts
chriswininger: hmm wasn't aware about that, I'm going to add a todo to explore this but don't want to change it so late in the game :-)

chriswininger: c1a61fc97c45b3739baad9f1ea5001c5180c5dac

chriswininger: sorry I was confused a9667ca2fbb9d356520a8fc135ab2da7b36bd830

MatthewGidcomb (approved):

PR #544, opened 2025-12-21
Partner improvements part 3: public-facing UI

MatthewGidcomb:
Further integration of partner information into the web portal UI.

Previously, we only displayed the partner logo on the public submission page. This PR moves the `partner` state to a Vuex store module, initialized during initial page render, to make it globally available, and adjusts the header to swap out the Infinite logo for the partner's logo.


web-portal/middleware/persistPartner.global.js
MatthewGidcomb, line 1: This is just a formatting fix

web-portal/store/partner.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+export const actions = {
+  LoadPartner: async (context) => {
+    const { query } = useRoute();
+    if (query.partner && query.partner !== context.state.name) {
+      const { data, error } = await useAsyncData('partner-fetch', () =>
+        useNuxtApp().$apiService.get(`/partners/name/${query.partner}`)
+        .catch(() => null) // eat 404s to avoid breaking everything
```
MatthewGidcomb, line 29: In the event of a malformed `partner` query param, we presumably want this to fail silently and display the Infinite logo as if unset.

web-portal/app.vue
```diff
@@ -6,3 +6,12 @@
...
 </template>
+
+<script setup>
+  import { useStore } from 'vuex'
+  const store = useStore();
+
+  // this is where the Nuxt docs say too put something like this
+  // https://nuxt.com/docs/4.x/getting-started/state-management#usage-with-pinia
+  await callOnce('partner', async () => store.dispatch('partner/LoadPartner'))
+</script>
```
MatthewGidcomb, line 17: This doesn't quite feel right, but it's what the docs say to do. This approach is specifically called out as a replacement for a Nuxt 2 feature built on top of Vuex to initialize store state on initial render.

I debated wrapping this in a composable called something like `useInitState`, but ultimately decided the extra indirection wasn't worth it unless/until there's something else to bootstrap here.

web-portal/components/Toolbar.vue
```diff
@@ -1,7 +1,12 @@
 <template>
   <div class="container">
     <div id="toolbar" style="z-index:21">
-      <ii-logo id="logo" icon-color="#fff" width="140" height="55" style="cursor: pointer" />
+      <template v-if="partnerLogo">
+        <img id="logo" :src="partnerLogo" :alt="`Brought to you by ${partnerName}`" height="55" />
+      </template>
+      <template v-else>
+        <ii-logo id="logo" icon-color="#fff" width="140" height="55" style="cursor: pointer" />
+      </template>
```
MatthewGidcomb: For discussion the next time we meet: the WRFL logo we used for the old feature is black-text-on-transparent-background, since it was displayed on the public submission form, but the header has a black gradient background. Not sure a single logo can support both cases.

web-portal/components/PartnerLogo.vue
chriswininger, line 1: yay sub-components :-) I always like breaking things up

web-portal/app.vue
chriswininger: yeah a little weird but sort of makes since :thinking:

api-server/test/events.authenticated.e2e-spec.ts
```diff
@@ -1329,12 +1448,14 @@ describe('Authenticated Events API', () => {
...
-        logo_url: 'https://example.com/partner-1.png',
+        light_logo_url: 'https://example.com/partner-1-light.png',
+      dark_logo_url: 'https://example.com/partner-1-dark.png',
       });
 
       const partner2 = await createPartner({
         name: 'Partner 2',
-        logo_url: 'https://example.com/partner-2.png',
+        light_logo_url: 'https://example.com/partner-2-light.png',
+      dark_logo_url: 'https://example.com/partner-2-dark.png',
```
chriswininger: these looks like they are formatted weird in github

api-server/README.md
```diff
@@ -71,6 +71,10 @@ $ npm run test:cov
 local: http://localhost:3003/api/
 prod:  https://api.infinite.industries/api/
 
+## Migrations
```
chriswininger, line 74: yay I finally added this, I swore I put this here like 5 times and some how never committed it

web-portal/store/partner.js
```diff
@@ -0,0 +1,38 @@
...
+      )
+      if (error.value) {
```
chriswininger, line 32: could put a new line above if

chriswininger (approved):
looks good to me

PR #545, opened 2026-01-04
Partner improvements: only show partner's own events if requested

MatthewGidcomb:
Pursuant to this morning's discussion, this PR adds support on the web portal for an `only` query param, which, when set to `partner` and paired with the `partner` param, will _only_ include that partner's events on the home page.

This filtering is implemented in the Vuex store, which makes the logic for requesting the events messier than I'd like, but we can consider moving it to the API later, once we've established this is behavior we event want.

I've also added logic on the home page (and the old `/remote` page, which we don't expose in the nav anymore) to suppress the "Online Resources / Projects" subheading if the result of filtering the remote events is an empty set.

### DIFF
--- a/web-portal/pages/index.vue
+++ b/web-portal/pages/index.vue
@@ -7,8 +7,10 @@
 
     <ii-list-viewer :calendar_events="events" />
 
-    <h2>Online Resources / Projects</h2>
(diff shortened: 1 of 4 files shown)


chriswininger (approved):
+1

PR #547, opened 2026-02-08
All partner admins to edit their events

chriswininger:
* Open access for partner-admins to edit their own events
* Corrects an issue in update endpoint where we did not return related models
* Provides tests
* Minor read me update


api-server/src/events/events.service.ts
```diff
@@ -279,44 +279,50 @@ export class EventsService {
     id: string,
     values: Partial<UpdateEventRequest>,
   ): Promise<DbUpdateResponse<EventModel>> {
```
chriswininger, line 278: The actual changes here are much much smaller than they appear, but I configured my IDE to apply our formatting rules on save and apparently this file was not following the rules. The big change is that we re-fetch the model (using the existing transaction). This makes sure we get all the related models.

```
   // refetching the event makes sure we get back all related models, since this is all inside a transaction it
      // should be safe
      const updatedEvent = await this.findOneWithRelated({
        where: { id },
        transaction,
      });
```

MatthewGidcomb (approved):

PR #552, opened 2026-03-13
Venue/Partner Association Part I

chriswininger:
* Creates an association table so that we can associate partners with venues.
* Updates database entities and api dtos to include partners
* Updates methods in events.service.ts so that we include the associated partners with the venues that are fetched on events.
* Updates venue.service.ts and related controller logic to fetch the related partner models
* Replaced instances where returned VenueModel directly over the API with a new VenueDTO. This is better from a separation of concerns stand point, but more importantly it avoided some really ugly TypeScript workarounds I was being forced to do :-)
* Updated related tests

This stops short of actually changing our permissions model. I will update isOwner in a followup PR. This is just putting the necessary building blocks in place.

### DIFF
--- a/api-server/src/authentication/authentication.controller.ts
+++ b/api-server/src/authentication/authentication.controller.ts
@@ -23,8 +23,7 @@ import {
   AUTH_USE_TEST_USERS,
 } from '../constants';
 import createJWTForTestUser from './utils/createJWTForTestUser';
-import UsersService from '../users/users.service';
-import { PartnersService } from '../users/partners.service';
+import { TestPartnerService } from './test-partner.service';
 
 @Controller(`${VERSION_1_URI}/authentication`)
 @ApiTags('authentication')
@@ -32,8 +31,7 @@ export class AuthenticationController implements OnModuleInit {
   constructor(
     @Inject(WINSTON_MODULE_NEST_PROVIDER)
     private readonly logger: LoggerService,
-    private readonly usersService: UsersService,
-    private readonly partnersService: PartnersService,
+    private readonly testPartnerService: TestPartnerService,
   ) {
     if (AUTH_USE_TEST_USERS) {
       this.logger.warn(AUTH_USE_TEST_USERS_WARNING);
@@ -42,7 +40,7 @@ export class AuthenticationController implements OnModuleInit {
 
   async onModuleInit() {
     if (AUTH_USE_TEST_USERS) {
-      await this.ensureTestPartnerUser();
+      await this.testPartnerService.ensureTestPartnerUser();
     }
   }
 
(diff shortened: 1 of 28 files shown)


api-server/src/authentication/authentication.controller.ts
```diff
@@ -142,53 +140,4 @@ export class AuthenticationController implements OnModuleInit {
       );
     }
   }
-
-  private async ensureTestPartnerUser() {
```
chriswininger, line 146: a little bit of cleanup, I shifted this to a separate file as Mathew originally suggested. I couldn't make it a simple utility function because of the need to handle dependency injection, so I created a proper service

api-server/src/authentication/test-partner.service.ts
```diff
@@ -0,0 +1,129 @@
...
+      );
+    } catch (error) {
+      if (error.message?.includes('already associated')) {
+        this.logger.log('User is already associated with partner');
+      } else {
+        throw error;
+      }
+    }
+
+    await this.ensureTestVenues(partner);
```
chriswininger, line 64: Most of this is the same as before, except for the addition of ensureTestVenues. This should help later with cypress tests. That won't come until at least PR part II :-)

api-server/src/events/dto/eventDTO.ts
```diff
@@ -109,8 +109,8 @@ export default class EventDTO {
   @ApiProperty({ example: [] })
   links: Array<string>;
 
-  @ApiProperty()
-  venue: VenueModel;
+  @ApiProperty({ type: () => VenueDTO })
+  venue: VenueDTO;
```
chriswininger, line 113: when ahead and created a proper VenueDTO, we were just returning VenueModel, but in addition to not being a good practice it was making the typescript really ugly due to weirdness in the Sequelize types

api-server/src/events/events.service.ts
```diff
@@ -232,11 +242,26 @@ export class EventsService {
             })
           : [];
 
+      // Derive venues and populate venue partners before the ownership check
+      // so that venue-partner associations are available when isOwner runs
       paginatedRows.forEach((event) => {
```
chriswininger: we're not updating the ownship yet, that will come in a follow up

api-server/test/venues.e2e-spec.ts
```diff
@@ -309,10 +378,20 @@ describe('Venues (e2e)', () => {
     }
   }
 
-  function deleteAllAnnouncements(): Promise<number> {
+  function deleteAllVenues(): Promise<number> {
```
chriswininger, line 381: whoops this was named wrong, copy/pasta :-)

api-server/src/events/dto/eventDTO.ts
MatthewGidcomb: This doesn't change any fields clients see, right?

api-server/src/events/events.service.ts
```diff
@@ -232,11 +242,26 @@ export class EventsService {
...
+          if (dt.venue && !uniqueVenues.has(dt.venue.id)) {
+            uniqueVenues.set(dt.venue.id, dt.venue);
+          }
+        });
+        event.venues = [...uniqueVenues.values()];
+      });
+
+      await this.populateVenuePartners(paginatedRows);
 
+      paginatedRows.forEach((event) => {
```
MatthewGidcomb, line 303: This may be an artifact of the way Github is showing me the diff, but it looks like this is an extra iteration over `paginatedRows`?

MatthewGidcomb: Could this be refactored to do everything in one pass through the paginated results, instead of (what looks like) 3?

api-server/test/current-events.e2e-spec.ts
```diff
@@ -386,6 +386,60 @@ describe('CurrentEvents (e2e)', () => {
       });
   });
 
+  it('returns venue partners on events when venues have associated partners', async () => {
```
MatthewGidcomb, line 389: What's the use case for returning the venue's partners on the current events endpoint?

api-server/src/events/dto/eventDTO.ts
chriswininger: no they should all be the same as of right now, though it gives us the option to change it in the future independent of the the DB

api-server/src/events/events.service.ts
chriswininger: yeah good call, I doubt the 3n matters much with hundreds of items in a page, but I think one pass might be easier to read too :-)

b21b3d795018bee7087300a1693e5cc84b6dca98

api-server/test/current-events.e2e-spec.ts
chriswininger: Just to be consistent, future proof, and make tests easier. I don't have a strong use case yet, though I'm planning to implement some admin ui interfaces soon that will almost definitely need it

PR #553, opened 2026-03-16
Venue Mapping Part-II: Ownship via Venue

chriswininger:
implement ownership logic and tests for ownership via venue

### DIFF
--- a/api-server/src/events/events.authenticated.controller.ts
+++ b/api-server/src/events/events.authenticated.controller.ts
@@ -46,7 +46,7 @@ import { validateAndExtractOptionalDateTimeFilters } from './utils/validateAndEx
 import { AuthenticatedUserGuard } from '../authentication/auth-guards/authenticated-user.guard';
 import { PartnerAdminGuard } from '../authentication/auth-guards/partner-admin.guard';
 import { RequestWithUserInfo } from '../users/dto/RequestWithUserInfo';
-import { Op } from 'sequelize';
+import { Op, literal } from 'sequelize';
 import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
 
 @Controller(`${VERSION_1_URI}/authenticated/events`)
@@ -209,13 +209,24 @@ export default class EventsAuthenticatedController {
 
     const partnerIds = user.partners?.map((partner) => partner.id) || [];
 
+    const partnerIdsList = partnerIds.map((id) => `'${id}'`).join(', ');
+
+    // It's safe to inject partnerIdsList because they are loaded from the database via the authenticated
+    // user's partner associations
     const findOptions = {
       ...getOptionsForEventsServiceFromEmbedsQueryParam(embed),
       where: {
         verified: false,
-        owning_partner_id: {
-          [Op.in]: partnerIds,
-        },
+        // We need to get any events that match owning_partner_id to on one of the users partner_ids or that
+        // take place at a venue that belongs to one of these partners
+        [Op.or]: [
+          { owning_partner_id: { [Op.in]: partnerIds } },
+          literal(`"EventModel"."id" IN (
+            SELECT dv.event_id FROM datetime_venue dv
+            JOIN venues_partners_mappings vpm ON vpm.venue_id = dv.venue_id
+            WHERE vpm.partner_id IN (${partnerIdsList})
+          )`),
+        ],
       },
     };
 
--- a/api-server/src/events/events.authenticated.controller.ts
+++ b/api-server/src/events/events.authenticated.controller.ts
@@ -2,7 +2,6 @@ import {
   Body,
   Controller,
   Delete,
-  ForbiddenException,
   Get,
   Inject,
   LoggerService,
@@ -175,12 +174,12 @@ export default class EventsAuthenticatedController {
   getAllNonVerified(
     @Req() request: RequestWithUserInfo,
   ): Promise<EventsResponse> {
-    const findOptions = {
-      where: { verified: false },
-    };
-
+    // Since infinite admins have access to all partners this will really just return
+    // all non-verified events
+    //
+    // We could consider a future enhancement to let admins filter by a given partner,
     return this.eventsService
-      .findAll(request, findOptions)
+      .findAllNonVerifiedForPartnersBelongToTheCurrentUser(request)
       .then((events) => events.map(eventModelToEventDTO))
       .then((events) => new EventsResponse({ events }));
   }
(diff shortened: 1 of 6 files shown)


api-server/src/events/events.service.ts
```diff
@@ -179,7 +179,7 @@ export class EventsService {
       // guaranteed to stay the same.
       // We've also added a secondary order by on created_at to ensure that events without start_time like
       // online resources at least sort consistently
-      const paginatedRows: EventModel[] = await this.sequelize.query(
+      const paginatedRows = (await this.sequelize.query(
```
chriswininger, line 228: This was a small tweak, unrelated to these changes that makes TypeScript happier

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -209,13 +209,24 @@ export default class EventsAuthenticatedController {
...
+        // We need to get any events that match owning_partner_id to on one of the users partner_ids or that
+        // take place at a venue that belongs to one of these partners
+        [Op.or]: [
+          { owning_partner_id: { [Op.in]: partnerIds } },
+          literal(`"EventModel"."id" IN (
+            SELECT dv.event_id FROM datetime_venue dv
+            JOIN venues_partners_mappings vpm ON vpm.venue_id = dv.venue_id
+            WHERE vpm.partner_id IN (${partnerIdsList})
+          )`),
+        ],
```
chriswininger: There may be a  better than falling back to raw-sql here, but this works and doesn't seem too bad.

e2e-tests/e2e/partnerWorkFlowsTest.js
```diff
@@ -88,8 +88,8 @@ context('Partner Work Flows', () => {
     const guid = crypto.randomUUID()
     const NON_PARTNER_EVENT_NAME = `Non-Partner Test Event-${guid}`
     
-    // Submit event without partner query parameter
-    submitEvent(NON_PARTNER_EVENT_NAME, PARTNER_ADMIN_USERNAME, PARTNER_ADMIN_PASSWORD, null)
+    // Submit event without partner query parameter, at a venue not associated with any partner
```
chriswininger, line 91: We were just picking the first venue which happened to also be the venue associated with the partner so it was breaking this existing test, this makes sure we pick a venue we know is not associated with the partner

api-server/src/events/events.authenticated.controller.ts
MatthewGidcomb: Indeed, that directly-injected array of partner IDs is scary even with the comment explaining the safety. Does that `literal` function support variables?

api-server/src/events/events.service.ts
```diff
@@ -649,11 +649,18 @@ export class EventsService {
       : null;
 
     const partnerIdsArray = ensureEmbedQueryStringIsArray(owningPartnerIds);
+    const partnerIdsList = partnerIdsArray
```
MatthewGidcomb: Is this list of partner IDs loaded from DB? `ensureEmbedQueryStringIsArray` seems to imply not.

api-server/src/events/events.authenticated.controller.ts
chriswininger: The tricky part is that the variables need to be passed in at the query level which isn't part of where we are constructing this, the way this is put together makes it clunky, that said I should probably just make a custom function for this in the service layer and thin I suspect maybe this gets easier.

I'm not super worried about because these are ids which come from the db and db has a schema there's not really a good way I can think of to insert something with a shape that would be exploitable into an id column, but maybe I'll poke at a refactor

chriswininger: I did shift the logic into the service layer but threading through the variables down the chain and getting them inserted correctly to make all this work without the direct injection feels super messy. Given this is all internal and coming from the DB I think we should just leave it how it is

api-server/src/events/events.service.ts
chriswininger: Yes, that is either a poorly named function or a poorly used function depending on how you think about it. They are passed into getWhereClause from the db. I'm going to look into why this method is even getting called, this feels really weird

chriswininger: or not  `@Query('owning_partner_id') owningPartnerIds: string[] | string = [],`, ugh maybe these aren't IDs, these are the partner names probably

chriswininger: ok that was a good catch, I forgot we had a place accepting these directly

chriswininger: This should take care of this plus tighten the instance where it comes from the DB for good measure b0f3e5796b495713e23959cd12dad1b81948be15

api-server/src/events/events.authenticated.controller.ts
chriswininger: After looking at your other comment, I was forced to do some necessary refactoring to anyways so this has all been tightened as part of that fix b0f3e5796b495713e23959cd12dad1b81948be15

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -175,12 +174,12 @@ export default class EventsAuthenticatedController {
   getAllNonVerified(
     @Req() request: RequestWithUserInfo,
   ): Promise<EventsResponse> {
-    const findOptions = {
-      where: { verified: false },
-    };
```
chriswininger, line 180: Shifted this into the service logic

api-server/src/events/events.authenticated.controller.ts
```diff
@@ -194,33 +193,10 @@ export default class EventsAuthenticatedController {
   @ApiResponse({ status: 403, description: 'Forbidden' })
   @ApiBearerAuth()
   async getAllNonVerifiedForPartnersBelongToTheCurrentUser(
-    @Query('embed') embed: string[] | string = [],
     @Req() request: RequestWithUserInfo,
   ): Promise<EventsResponse> {
-    const user = request.userInformation;
```
chriswininger, line 200: Shifted all this into service logic

api-server/src/events/events.service.ts
```diff
@@ -164,22 +198,34 @@ export class EventsService {
     }
     return this.sequelize.transaction(async () => {
       const tagClauseParams = this.getTagsClauseParams(tags);
-      const whereClause = this.getWhereClause(
-        tagClauseParams,
-        category,
-        owningPartnerIds,
-        verifiedOnly,
+      const { whereClause, replacements: whereClauseReplacements } =
```
chriswininger, line 201: We're making sure everything is handled via replacement in SQL so the dynamic where clause now had to return a set of dynamic variables for the replacement

api-server/src/events/events.service.ts
```diff
@@ -639,22 +696,32 @@ export class EventsService {
...
+      replacements.filterPartnerIdsOwning = partnerIdsArray;
+      replacements.filterPartnerIdsVenue = partnerIdsArray;
+      partnerIdsClause = `(events.owning_partner_id IN (:filterPartnerIdsOwning)
+           OR EXISTS (
+             SELECT 1 FROM datetime_venue dv2
+             JOIN venues_partners_mappings vpm ON vpm.venue_id = dv2.venue_id
+             WHERE dv2.event_id = events.id
+             AND vpm.partner_id IN (:filterPartnerIdsVenue)
+           ))`;
+    }
```
MatthewGidcomb, line 724: More of a question than a comment: is it strictly necessary to do two separate replacements for the partner IDs, or just convention / future-proofing in case we have some situation where they might become different sets?

MatthewGidcomb (approved):

PR #554, opened 2026-03-23
Node, Nuxt, and various web-portal upgrades and cleanup

MatthewGidcomb:
This is an experimental upgrade of Nuxt to latest v3, `3.21.2`, and also bumps the Node version in the top-level `.nvmrc` file because Node 20 reaches end-of-life in a few weeks.

While there are no breaking changes in the Nuxt versions, I did make some code updates based on the release notes: some fields on the internal error-handling are changing in Nuxt 4, and the change has been backported to Node 3 with the old field names marked as deprecated.

**Update**: I've perhaps gone a bit overboard with additional dependency upgrades, doing the `@fullcalendar/` ecosystem, `moment-timezone` (pulls in updates to the timezone db), `nuxt-auth-utils`, `winston`, and a few others, and updated a deprecated CSS syntax that throws a lot of warnings in the dev server console.

Things we'll need to verify when we take this to staging:

- Does auth work correctly?
- Are dates rendered correctly, in the correct timezone?
- Does the calendar page work correctly? (there are some known styling issues with the previous/next buttons, don't panic on that)
- Does the nav look right? (one of the two places where I replaced the deprecated `>>>` with `:deep(...)`)
- Do the event tables in admin look right? (the other)


web-portal/components/AdminEventsList.vue
```diff
@@ -67,17 +67,21 @@
     font-family: "Open Sans", sans-serif;
   }
 
-  .calendar-events-table >>> td, .calendar-events-table th {
+  .calendar-events-table :deep(td), .calendar-events-table th {
```
MatthewGidcomb, line 70: For reference, this is the syntax for breaking out of (Vue's) CSS-scoping, allowing this rule to target an element in a descendant component (in this case, the `ii-admin-events-list-row` component).

.nvmrc
```diff
@@ -1,2 +1 @@
-lts/iron
-
+lts/jod
```
chriswininger, line 1: If we're going to bump this we should probably also bump it in these spots https://github.com/infinite-industries/infinite/pull/555

chriswininger (approved):

.nvmrc
MatthewGidcomb: Good catch, I forgot about the pipelines.

PR #558, opened 2026-05-02
Fix faulty isDirty function

summasmiff:
## Bug Description
When the Submit Event form is navigated away from, the user is told they have unsaved changes and asked if they are sure. This happens on every navigation away from the submit form, even when the form is completely blank. 

## Root Cause
`isDirty` was a check intended to return `true` if there were local unsaved changes and `false` if there were not. Due to formatting applied in the `created` hook of the submit form, this function always returned true because the form version of the event was different from the server version of the event.

## Solution
Fixed by applying the same formatting to the server version of the event in the `isDirty` check so the only changes detected would be actual unsaved user changes.

### DIFF
--- a/web-portal/components/SubmissionForm.vue
+++ b/web-portal/components/SubmissionForm.vue
@@ -393,6 +393,12 @@
     }
   })
 
+  const createFormEvent = (event) => Object.assign({}, event, {
+    condition: event.condition ? event.condition.map(t => t) : [],
+    date_times: event.date_times ? event.date_times.map(dt => ({ ...dt })) : [],
+    tags: event.tags ? event.tags.map(t => t) : []
+  })
+
   export default {
     props: ['event_id', 'user_role', 'user_action', 'owningPartnerId'],
     // user_role --> admin, venue, regular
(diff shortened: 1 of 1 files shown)


MatthewGidcomb (approved):
I'm a bit confused as to _why_ this works, but it definitely does. Thanks for fixing this.

summasmiff:
> I'm a bit confused as to _why_ this works, but it definitely does. Thanks for fixing this.

To diagnose this, I logged in the isEqual check exactly which fields were considered not equal between an empty form event and the empty stored event. It flagged the `event.condition` which in the form was set to `[]`. 

In the `created:` step, we explicitly assign the event `condition`, `date_times` and `tags` to an empty array on page load if it is false-y, which would make the form version different from the server version according to the isEqual check.

false-y in JS would be undefined, null, or false, so that would not equal `[]`.

web-portal/components/SubmissionForm.vue
```diff
@@ -425,19 +431,16 @@
     },
     created: function () {
       const new_event = this.$store.getters.GetCurrentEvent
-      this.calendar_event = Object.assign({}, new_event, {
-        condition: new_event.condition ? new_event.condition.map(t => t) : [],
-        date_times: new_event.date_times.map(dt => ({ ...dt })),
-        tags: new_event.tags ? new_event.tags.map(t => t) : []
-      })
+      this.calendar_event = createFormEvent(new_event);
```
chriswininger, line 434: thx for the readability clean up :-)

chriswininger (approved):
nice 👍

PR #559, opened 2026-05-03
Fix incorrect field in Resource Template Service, and some cleanup

MatthewGidcomb:
Follow-up to #558, where after discussion this morning I realized that the create-new-blank-event helper was also initializing the `condition` field to empty string instead of empty array (note: this field maps to a Postgres array-of-string column, and can contain the values `postponed`, `cancelled`, and `sold-out`).

This PR makes that fix, and also reorders the fields in the return value to roughly match the form itself (helpful in determining that all fields were accounted for and of the correct type), removes some that are no longer relevant, and updates the doc string.

I've also removed two methods on the `SubmissionForm` component that referenced the deprecated-and-removed field `additional_dates`, and added two test cases for the component's `isDirty` method (they're not _great_ tests, but would have helped us catch #558 earlier).

### DIFF
--- a/web-portal/test/SubmissionForm.spec.js
+++ b/web-portal/test/SubmissionForm.spec.js
@@ -6,11 +6,6 @@ import { getEmptyCalendarEvent } from '../services/ResourceTemplateService'
 
 vi.mock('@/services/ImageUploadService')
 
-/**
- * When submission was refactored (Nov 2021) the only test case here
- * was rendered irrelevant. Keeping this for reference for future testing
- * of this component.
- */
 describe('SubmissionForm component', () => {
   let store = null
   let wrapper = null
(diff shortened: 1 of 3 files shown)


web-portal/test/SubmissionForm.spec.js
```diff
@@ -184,6 +179,15 @@ describe('SubmissionForm component', () => {
     expect(wrapper.vm.calendar_event.condition).toEqual(['sold-out'])
   })
 
+  it('isDirty method returns false if no edits have been made', () => {
+    expect(wrapper.vm.isDirty()).toBe(false)
+  });
```
summasmiff, line 184: Thanks for these!

summasmiff (approved):

PR #561, opened 2026-05-08
Add local dev setup script and docs

summasmiff:
Adds `bin/setup.sh` to start a local development environment:

- Spins up Docker container for database (restarts if it exists)
- Waits for DB & runs migrations
- Starts API server & web portal in background
- Auto-opens browser when ready

Also adds DEVELOPING.MD to the root level of the repo with some README type details - I found the 2 READMEs in /api_server and /web_portal a little out of date, maybe we can look at them together soon.

### DIFF
--- a/DEVELOPING.MD
+++ b/DEVELOPING.MD
@@ -1,0 +1,32 @@
+# Developing Infinite Industries
+The Infinite Industries web app consists of the frontend `./web-portal`  built using Vue & Nuxt and backend `./api-server` built using NestJS. For more info on each, please see their respective READMEs.
+
+## Tech stack
+### Database
+- Postgres
+
+### API Server
+- [NestJS](https://docs.nestjs.com/): API framework for NodeJS
+
+### Web Portal
+- [Nuxt 3](https://nuxt.com/): Framework for Vue with opinionated defaults
+- [Vue](https://vuejs.org/guide/introduction.html): JS library for building reactive web apps
+
+## Local dev startup
+There are a few ways to start the app for local development. One way:
+
+ `$ ./bin/setup`
+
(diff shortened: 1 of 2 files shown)


DEVELOPING.MD
MatthewGidcomb, line 1: Would be nice to get a reference in here to the dev container setup, but maybe after we've fixed it...

MatthewGidcomb:
Might be worth referencing the `DEVELOPING` doc from the main README. Sounds like Chris and I should also update the component-level READMEs.

DEVELOPING.MD
chriswininger: yeah last time I tried to use it was kind of rough. I got it to work by making some changes but I didn't check them in because I wasn't very confident in what they were doing and to what extent it was changes to make it work on my machine vs on anyone's

chriswininger (approved):
Nice thanks for all the documentation

bin/setup.sh
```diff
@@ -0,0 +1,138 @@
...
+    echo "Starting web portal..."
+    npm run start:dev
+) &
+
+wait_and_open_browser &
+
+echo "All services started. Press Ctrl+C to stop."
+
+# Keep script running to call cleanup
+wait
```
chriswininger, line 138: Some lovely bash, +1 for thinking through the cleanup. This all seems nice and readable or at least as readable as bash can be :-)

chriswininger:
@summasmiff Merge when ready +1

DEVELOPING.MD
summasmiff: Happy to continue to grow this file but wanted this first draft to only include things I was 100% confident of :)

PR #564, opened 2026-05-19
Change copy link button to use full URL

### LINKED ISSUE #465: add a copy to our share options
if you are viewing the site in app mode, by adding it to your home screen on ios there is no easy way to copy urls. Many apps deal with this via copy button under the share options

MatthewGidcomb:
We used to generate a Bitly short link for events as part of the creation process, but stopped quite a while ago. The button on the individual event view (in the Share menu) that supported copying this short link to the user's clipboard is hidden if it wasn't set on the event.

This PR repurposes that button / menu item for copying the full canonical URL for the event to the user's clipboard.

Resolves #465

### DIFF
--- a/web-portal/components/EventView.vue
+++ b/web-portal/components/EventView.vue
@@ -92,9 +92,8 @@
                 <span>Tweet</span>
               </a>
               <div
-                v-if="event.bitly_link"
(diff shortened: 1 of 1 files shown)


summasmiff (approved):
Nice fix!

PR #569, opened 2026-06-07
Deploy 2026-06-07

MatthewGidcomb:
- Initial partner management UI (admin)
- Fix for `isDirty` checks on submission form
- Fix for layout issue on submission form
- Restore old copy Bitly button, repurposed for copying the full URL instead
- Various docs updates and setup scripts


summasmiff (approved):
*high fives self*

PR #574, opened 2026-06-29
Show different datepicker for Single Day, Multi-day, and No Date events (eg online only)

summasmiff:
## Goals of this PR
- Show separate datepickers for different event types
- Refactor SubmissionForm & child components into more modular components for easier state management and future improvements

## New datepickers
New datepicker states are based on [these Figma mockups](https://www.figma.com/proto/TYVAMSagx0ufI1DnT9A7ZZ/Infinite-Industries-Submission-Form-2?node-id=169-809&p=f&t=8fPm2CGnhw9vFxIM-1&scaling=min-zoom&content-scaling=fixed&page-id=3%3A8&starting-point-node-id=169%3A809&show-proto-sidebar=1). 

### - No Date (Online Resource)
Hides the picker completely.

### - Single Date (Single-day event, Call for Entry)
Renders one DateTimeVenueEditor.
Screenshot 2026-07-14 at 3 07 16 PM

### - Multi-Day (Gallery Show, Multi-day Event, Other)
Renders one DateTimeVenueEditor. Once a first date is selected, shows a button to add more.
Screenshot 2026-07-14 at 3 08 13 PM

## New "Add Venue" behavior
The "Add Venue" form is now less prominent to prevent unnecessary venue additions. It is available from the Search Venues component as the final option, "Add a new venue...". When clicked, this opens a modal.
Screenshot 2026-07-14 at 3 10 27 PM
Screenshot 2026-07-14 at 3 10 39 PM

## Refactor notes: new components: EventDateTimesVenuesEditor & DateTimeVenueEditor
The purpose of this PR is to add better UX for different types of events. Adding a bunch of new state management into the already-long SubmissionForm felt like a step backwards. Instead, this PR adds a couple new components to manage the local concern of "which datepicker to show". 

#### EventDateTimesVenuesEditor
The main smart container managing the DateTimeVenues array (v-model="calendar_event.date_times"). It handles the conditional logic for three states, Single Day, Multi-Day, and No Date.

#### DateTimeVenueEditor
A 1:1 dumb presentation component combining the time and venue pickers. Manages confirmation/edit states and toggles between editable or static text displays.

Updates have been made to DateTimePicker and VenuePicker to make them "dumber" (more composable).


node_modules/.package-lock.json
MatthewGidcomb, line 1: This looks like an artifact of an accidental top-level `npm install` run?

web-portal/components/AddNewVenue.vue
```diff
@@ -1,205 +1,429 @@
...
+          <li
+            v-for="(city, index) in filteredCities"
+            :key="city"
+            :class="{ 'active': index === cityHighlightIndex }"
+            @mousedown.prevent="selectCity(city)"
+            @mouseenter="cityHighlightIndex = index"
+          >
+            {{ city }}
+          </li>
+        </ul>
```
MatthewGidcomb: I don't mind dropping Vuetify components here, but do we need any `aria-` props to make sure this is at least marginally accessible?

MatthewGidcomb: (I see the keyboard handling, which looks good, but wondering if we need `aria-haspopup` and/or `aria-expanded`/`aria-controls`)

web-portal/components/DateTimeVenueEditor.vue
```diff
@@ -0,0 +1,294 @@
...
+    </div>
+    <div class="confirmed-entry-actions">
+      <button type="button" class="delete-entry-button" @click="deleteEntry">&times;</button>
+    </div>
+  </div>
+</template>
+
+<script setup>
+  import { computed, nextTick, ref, watch } from 'vue'
+  import Edit from './vectors/Edit.vue'
```
MatthewGidcomb: Nitpick but for consistency the import path should probably be `'@/components/vectors/Edit.vue'`

web-portal/components/DateTimeVenueEditor.vue
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  const handleNewVenue = (venue) => {
+    selectedVenueId.value = venue?.id
+    emit('selectVenue', venue)
+  }
+</script>
+
+<style scoped>
+.editable-form {
```
MatthewGidcomb: Another nitpick, but stylistically I like to have the `<style>` blocks indented one level (2 spaces), just for consistency with the `<template>` and `<script>`.

web-portal/components/EventDateTimesVenuesEditor.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <div class="datetime-venue-list">
+    <!-- Confirmed DatetimeVenues -->
+    <div v-if="localEntries.length > 0" class="confirmed-entries">
+      <h4>When and Where:</h4>
+        <date-time-venue-editor
+          v-for="(entry, index) in localEntries"
+          :key="entry.id || index"
+          v-model="localEntries[index]"
```
MatthewGidcomb, line 9: Can this be `v-model="entry"`?

web-portal/components/TimePicker.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <span class="time-part">
+    <input
+      :ref="hourRef"
+      v-model="hourLocal"
+      type="text"
+      placeholder="hh"
+      maxlength="2"
+      :class="invalidHour ? 'invalid hour-part' : 'hour-part'"
```
MatthewGidcomb: This is fine but it's also possible to mix bound and unbound `class` (see https://vuejs.org/guide/essentials/class-and-style.html)

```vue
<input
  // ...
  class="hour-part"
  :class="{ invalid: invalidHour }"
/>
```

package-lock.json
MatthewGidcomb, line 1: Accidental?

web-portal/components/EventDateTimesVenuesEditor.vue
summasmiff: No, because that's not the actual state object, it's just the read-only reference for the `v-for` (used in the key). The `localEntries[index]` is the actual object to be manipulated - since v-model is both read AND write it needs to access the "official" version.

web-portal/components/AddNewVenue.vue
summasmiff: This has been updated to make both the City and State inputs into proper "combobox" elements: 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role

- Input element is a combobox and has `aria-autocomplete`, `aria-expanded`, `aria-haspopup` and `aria-activedescendant` attributes for identifying, showing, selecting, and hiding the dropdowns
- Dropdowns are selectable by keyboard and screenreader as well as by mouse

Furthermore, the entire form is now wrapped in a `<form>` element so it can be submitted by a screenreader.

web-portal/components/vectors/Edit.vue
MatthewGidcomb, line 1: I like the new icon.

This may be more trouble to actually do than it's worth (I'm definitely not an SVG artist), but I wonder if a downward-facing curved gap near the top, about where the bottom of the metal part that holds the eraser on a pencil would be, would give it a little more sense of volume?

MatthewGidcomb: _Very_ rough drawing-on-a-screenshot that probably illustrates it better than that description:

Image

web-portal/components/AddNewVenue.vue
```diff
@@ -1,205 +1,621 @@
...
+        <div class="form-group autocomplete-wrapper">
+          <label id="venueCityLabel" for="venueCity">
+            City<span class="sr-only"> (required)</span><span class="required-field">*</span>
+          </label>
+          <div class="autocomplete-container" ref="cityDropdownContainer">
+            <input
+              type="text"
+              id="venueCity"
+              role="combobox"
+              aria-autocomplete="list"
```
MatthewGidcomb, line 42: My only concern on the ARIA roles is that the MDN docs and spec say not to use `aria-autocomplete` if the suggestions aren't dependent on the user's input. That's sometimes-but-not-always the case here: the list filters as you type but has the initial list of suggestions with no search input.

Might be better to err on the side of not setting `aria-autocomplete`.

MatthewGidcomb: With the caveat that I don't have a lot of experience setting ARIA attributes manually, the other settings on this component look right to me.

MatthewGidcomb: And now that I've got my local setup working and can experiment with the PR locally, the keyboard navigation feels right to me, too.

Except for one thing: tabbing out of one of the combobox fields focuses the listbox, and tabbing again to focus the next field doesn't hide the listbox.

web-portal/components/TimePicker.vue
```diff
@@ -0,0 +1,63 @@
...
+    <input
+      :ref="hourRef"
+      v-model="hourLocal"
+      type="text"
+      placeholder="hh"
+      maxlength="2"
+      class="hour-part"
+      :class="{ invalid: invalidHour }"
+    />
+    <span>:</span>
```
MatthewGidcomb, line 12: As-is this matches the current appearance, but experimenting with it I wonder if it would look better if this `:` had a little space around it, with either a space on each side or a small inline margin on the span...

web-portal/components/VenueSearch.vue
```diff
@@ -0,0 +1,212 @@
...
+        <p>{{ venue.address }}</p>
+      </div>
+
+      <!-- Always-visible "Add New" Option -->
+      <div
+        class="add-new-option"
+        @mousedown="openNewVenueModal"
+      >
+        + Add new venue...
+      </div>
```
MatthewGidcomb, line 31: One thing I like about this new approach is that you don't see it until you've searched down to a few result or scroll all the way to the end of the results. Hopefully this cuts down on duplicate venue submissions.

web-portal/components/DateTimePicker.vue
```diff
@@ -1,154 +1,42 @@
...
-
-          <div v-show="!picker&&!introduction">
-            If you need additional dates, please select them now.
-          </div>
-          <!-- <div v-show="!picker&&!introduction">
-              It's lonely here.
-            </div> -->
+    <div class="time-date-input-box">
+      <div class="time-date-entry">
+        On <button id="date-picker" class="link-button" @click="toggleCalendarModal()">{{ picker }}</button> from
```
MatthewGidcomb, line 5: This link-style button matches the design, but seeing it in action I still wonder if it would be clearer as an input that opens the picker on focus (which is how Flatpickr works out of the box, our wrapper sets `inline: true` to get the inline behavior).

chriswininger:
image

Just capturing what we discussed here to have it in writing. I think the link format should be an input, I'm not fully sold per-populating the time, but more open on that.

Probably sold on this, but at a first glance I was unsure how to find add new venue, maybe a good thing, just calling it out

image

web-portal/components/AddNewVenue.vue
```diff
@@ -1,205 +1,629 @@
...
+        <span class="sr-only">Saving venue...</span>
+      </div>
+    </template>
+
+  </Modal>
 </template>
 
-<script>
+<script setup>
+  import { ref, computed, onMounted, onBeforeUnmount, defineExpose } from 'vue'
```
MatthewGidcomb: Pulled the branch down locally to experiment, and I'm seeing a build warning that `defineExpose` is a compiler macro and doesn't need to be imported.

This must be true (that it doesn't need to be imported) because `VenueSearch` uses it without an import. I'm not a fan of magic like this but warnings in the built output are annoying.

web-portal/components/DateTimePicker.vue
```diff
@@ -1,154 +1,42 @@
...
+          hour-ref="endHourInput"
+          min-ref="endMinInput"
+          ampm-ref="endAmPm"
+        />
+        <select ref="eventTimezone" name="event_timezone" v-model="event_timezone">
+          <option v-for="(tz) in $config.public.timezoneOptions.split(',')" :key="tz">
+            {{ tz }}
+          </option>
+        </select>
+        <div v-if="show_calendar_modal" :key="picker || 'empty'">
```
MatthewGidcomb, line 35: Not sure why I don't see any space at all between the end time meridiem select and the timezone; seems like there's a little space between them in your screenshot.

Image

Probably needs at least the equivalent of a space char between them?

web-portal/components/DateTimePicker.vue
```diff
@@ -184,38 +63,58 @@
...
+        const startTime = currentTime.clone().minute(0).second(0);
+        const endTime = startTime.clone().add(1, 'hour').minute(0).second(0);
+
+        this.picker = currentTime.format('YYYY-MM-DD');
+        this.start_hour = '';
+        this.start_minute = startTime.format('mm');
+        this.start_ampm = startTime.format('a');
+        this.end_hour = '';
+        this.end_minute = endTime.format('mm');
+        this.end_ampm = endTime.format('a');
```
MatthewGidcomb: I like where we landed on the defaults here, as the minutes defaulted to `00` is basically equivalent to the current behavior but more explicit.

Two minor follow-up thoughts:

- Would it be clearer to set to minutes to `'00'` explicitly, rather than the minutes from the generated start/end times?
- Would it make sense to default AM/PM to `'PM'`, since most of our events are afternoon/evening?

MatthewGidcomb (approved):

